PR #128, opened 2023-06-08
FCS_CKM_EXT.8 change

### LINKED ISSUE #124: PBKDF from FDE proposal
The CCDB Crypto Working group struggled with this SFR for a while and was working with the DSC iTC to figure out a solution. This is what the WG currently is proposing:

FCS_CKM_EXT.8.1 The TSF shall perform password-based key derivation functions in accordance with a specified cryptographic algorithm [HMAC-[selection: SHA-256, SHA-384, SHA-512]], with iteration count of [assignment: number of iterations] using a randomly generated salt of length [assignment: length of salt] and output cryptographic key sizes [selection: 128, 192, 256] bits that meet the following standard: [NIST SP 800-132].

There's a lot of guidance text, but we wanted to keep the SFR focused on aspects of the standard (which only talks about number of iterations and length of the salt), rather than include aspects out side of the standard (e.g., anti-hammering).

-----

This is related to [DSC0001](https://dsc-itc.github.io/TD/DSC0001.html) and would be a further change in v1.1.

woodbe:
This is to resolve #124.

I have added FCS_CKM_EXT.8 according to the recommendation from @jdonndelinger

To resolve the issue about hammering I added a simplified version of FIA_TRT_EXT.1 from the PP_MDF that just specifies that a max attempts allowed of 10 per 500ms, but without any specific options for how this would be done.

I think I have all the dependencies and such properly added here. The SD update for the FIA_TRT_EXT.1 would largely be what is in the PP_MDF. We would still need the info for the new crypto SFR though as it seems likely to be different than FCS_COP.1/PBKDF (though maybe not too much).

### DIFF
--- a/cPP_DSC.adoc
+++ b/cPP_DSC.adoc
@@ -24,6 +24,8 @@ This collaborative Protection Profile (cPP) was developed by the {iTC-longame} i
 [.text-center]
 Apple Inc.
 [.text-center]
+Google, LLC
+[.text-center]
 Samsung Electronics Co., Ltd.
 [.text-center]
 *Common Criteria Test Laboratories*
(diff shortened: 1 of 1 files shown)


woodbe:
FIA_TRT_EXT.1 could become FIA_AFL_EXT.1.x

Main questions to review are limits on the numbers, and are they relevant.

Current proposal has this as limiting all attempts (successful or failed).

woodbe:
I am recommending to close this pull request, though I want to consider the options regarding throttling or binding or something along those lines. I think this particular change though can be deleted in favor of the crypto update and figure out what, if anything, is needed in addition.

PR #144, opened 2023-08-14
make AKs optional

### LINKED ISSUE #38: CCDB AU/008 - Section 6.3.15
FCS_STG_EXT.1.1

### Comment:
Storage for asymmetric private keys is not optional whereas other places in the PP allow asymmetric cryptography to be optional.  For example, the SFR  FCS_CKM.1.1/AK is selection based,  It is a reasonable use case to use a DSC for symmetric cryptography only.

### Proposal:
Consider modifying FCS_STG_EXT.1.1 to have public key storage as an optional selection.

### Priority: 1

woodbe:
This is to make AKs optional for the protected storage. Removes the "no other key" option since at least one key type (or secrets) needs to be selected.

This is to close #38.

The one thing about this specific change is that it could leave you with "persistent secrets" as the selection option, and I'm not sure that is what we expect (though I don't know if this would be an issue if someone could justify it). If we don't want to allow that, then I would probably break this into two selections, one for the key types (AK/SK) and then another one (persistent secrets/no other). That is a little more complicated, technically, but would be clear about expectations.


woodbe:
Merge based on call discussion

PR #145, opened 2023-08-14
debug modes

### LINKED ISSUE #39: CCDB AU/009 - Section 6.7.2
FPT_MOD_EXT.1

### Comment:
Debug modes are very useful in many situations when system integration does not work as it should.  Not allowing debug is only suitable in use-cases where the DSC is embedded in a larger system also developed by the DSC vendor.  In cases where a DSC is used as a component in some other vendors product a debug mode would be very useful, 

### Proposal:
Consider making FPT_MOD_EXT.1 optional.

### Priority: 2

woodbe:
app note update

This is to close #39.


woodbe:
merged on call

PR #151, opened 2023-09-12
update o.fw_integrity

### LINKED ISSUE #7: atsec/7 - Section 5.3 - incorrect mapping in table 2
Table 2

### Comment:
Table 2: a value in the table is inconsistent with CC v3.1 Part 1 (CC-1).  In Table 2, an assumption of Operational Environment (OE), A.ROT_INTEGRITY, is addressed by a security objective for TOE, O.FW_INTEGRITY.  As per CC-1, the security objectives for TOE do not trace back to assumptions to OE.

### Proposal:
Drop this mapping

### Priority: 1

woodbe:
this is to close #7.


woodbe:
Merged based on call agreement

PR #156, opened 2023-09-28
Key storage updates

### LINKED ISSUE #153: [cPP CHANGE] FCS_STG_EXT.2 & 3
FCS_STG_EXT.2 refers to FCS_COP.1/KeyEnc, which is no longer a thing.

I've been trying to figure out a way to get rid of FCS_STG_EXT.2 & 3 and replace them with FCS_CKM_EXT.3. I haven't quite gotten there, but I suggest the following change:

FCS_STG_EXT.2.1. The TSF shall encrypt [AKs, SKs, KEKs, and [selection: long-term trusted channel key material, all software-based key storage, no other keys]] using one of the following methods: [selection: key encapsulation, key wrapping].

There are a couple problems:
At this point key wrapping does not include GCM and CCM modes, but maybe the CWG will include them. This would also eliminate the need for FCS_STG_EXT.3, since all those methods provide integrity as well as confidentiality.

The other problem is that the old FCS_COP.1/KeyEnc allows "encryption" through XOR of a key with a KEK. This isn't allowed by the new catalog. Nor does it provide integrity. It would make things a lot simpler if we could get rid of  XOR because then all the allowed methods would include integrity and we could get rid of STG_EXT.3.

woodbe:
This is to close #153.

### DIFF
--- a/cPP_DSC.adoc
+++ b/cPP_DSC.adoc
@@ -1409,6 +1409,22 @@ FCS_COP.1.1/SKC:: The TSF shall perform _symmetric-key encryption/decryption_ in
 
 {empty}[*selection*: [.underline]#ISO/IEC 10116:2017 (Clause 7), NIST SP 800-38A#] [CBC]
 
+|AES-CCM
+|AES in CCM mode with unpredictable, non-repeating nonce, minimum size of 64 bits
+|{empty}[*selection*: [.underline]#128 bits, 192 bits, 256 bits#]
+|{empty}[*selection*: [.underline]#ISO/IEC 18033-3 (Sub Clause 5.2), FIPS PUB 197]# [AES]
+
+{empty}[*selection*: [.underline]#ISO/IEC 19772:2020 (Clause 7), NIST SP 800-38C]# [CCM]
+
+|AES-GCM
+|AES in GCM mode with non-repeating IVs
+
+IV length must be equal to 96 bits; the deterministic IV construction method [SP800-38D, Section 8.2.1] must be used; the MAC length t must be one of the values 96, 104, 112, 120, and 128 bits.
+|{empty}[*selection*: [.underline]#128 bits, 192 bits, 256 bits#]
+|{empty}[*selection*: [.underline]#ISO/IEC 18033-3 (Sub Clause 5.2), FIPS PUB 197]# [AES] 
+
+{empty}[*selection*: [.underline]#ISO/IEC 19772:2020 (Clause 10), NIST SP 800-38D]# [GCM]
+
 |XTS-AES
 |AES in XTS mode with unique [*selection*: [.underline]#consecutive non-negative integers starting at an arbitrary non-negative integer, data unit sequence numbers#] tweak values
 |{empty}[*selection*: [.underline]#256 bits, 512 bits#]
@@ -1429,6 +1445,22 @@ FCS_COP.1.1/SKC:: The TSF shall perform _symmetric-key encryption/decryption_ in
 |ISO/IEC 18033-3:2010 (Sub Clause 5.3) [Camellia]
 
 {empty}[*selection*: [.underline]#ISO/IEC 10116:2017 (Clause 7), NIST SP 800-38A#] [CBC]
+
+|CAM-CCM
+|Camellia in CCM mode with unpredictable, non-repeating nonce, minimum size of 64 bits
+|{empty}[*selection*: [.underline]#128 bits, 256 bits#]
+|ISO/IEC 18033-3:2010 (Sub Clause 5.3) [Camellia]
+
+{empty}[*selection*: [.underline]#ISO/IEC 19772:2020 (Clause 7), NIST SP 800-38C#] [CCM]
+
+|CAM-GCM
+|Camellia in GCM mode with non-repeating IVs
+
+the IV length must be equal to 96 bits; the deterministic IV construction method [SP800-38D, Section 8.2.1] must be used; the MAC length t must be one of the values 96, 104, 112, 120, and 128 bits.
+|{empty}[*selection*: [.underline]#128 bits, 256 bits#]
+|ISO/IEC 18033-3:2010 (Sub Clause 5.3) [Camellia]
+
+{empty}[*selection*: [.underline]#ISO/IEC 19772:2020 (Clause 10), NIST SP 800-38D#] [GCM]
 
 |XTS-CAM
 |Camellia in XTS mode with unique [selection: consecutive non-negative integers starting at an arbitrary non-negative integer, data unit sequence numbers] tweak values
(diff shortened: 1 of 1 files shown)


cPP_DSC.adoc
```diff
@@ -1023,6 +1023,94 @@ FCS_CKM.2.1:: The TSF shall distribute cryptographic keys in accordance with a s
...
++
+*_Cryptographic key usage in cryptographic operations* - Depending on the application, a cryptographic key could be root key which by virtue of being a root key, cannot be encrypted with another key.  It is often stored in plain sight, sometimes with hardware protections, sometimes obfuscated in clever ways, such as in physically unclonable functions.  Other than root keys, it is expected that other cryptographic keys should be protected with encryption, often using other keys, which leads to a chain or hierarchy of cryptographic keys that encrypt other keys often anchored by a root key.   Additionally, it is expected that an entity authenticate itself self as an authorized user of the key.  This latter requirement is in addition to the recommended requirements of the Common Criteria Part 2 for FCS_CKM.3._
```
smuellerDD: editorial: ... itself self ...

smuellerDD: Regarding the last sentence: What exactly should that statement imply - any special actions for the ST author?

cPP_DSC.adoc
```diff
@@ -1023,6 +1023,94 @@ FCS_CKM.2.1:: The TSF shall distribute cryptographic keys in accordance with a s
...
++
+*_Cryptographic key recovery* - This refers to the retrieval of cryptographic keys from either archival, backup, or escrow locations.  In each case, the TOE uses the agreed upon cryptographic key access method agreed upon._
++
+*_Modifications to attributes of cryptographic keys* - Often during the creation of cryptographic keys, the TOE assigns certain attributes to them.  This may include allowed key access methods, authentication methods, and other items as discussed above in Creation of cryptographic keys.  The TOE may require authentication for authorization to modify these attributes._
++
+*_Cryptographic key destruction* - The attributes of the cryptographic key should indicate the necessary conditions needed to destroy it.  Often this will be an authentication indicating authorization to destroy it.  This SFR would apply to keys stored in the TOE, and FCS_CKM.6 would apply for destruction.  Otherwise, the destruction of keys stored outside the TOE is not enforceable by this SFR._
```
smuellerDD: editorial: ... FCS_CKM.6 would apply ...
-> shall apply ?

cPP_DSC.adoc
```diff
@@ -1023,6 +1023,94 @@ FCS_CKM.2.1:: The TSF shall distribute cryptographic keys in accordance with a s
...
++ 
+*_Cryptographic key archival, backup, and escrow* - TOEs often perform cryptographic key archival to manage limited memory resources inside its own security boundaries.  Often the TOE encrypts the cryptographic keys prior to saving them to storage that is close by on the same device as the TOE and that can be accessed quickly. TOEs may perform cryptographic key backups into storage that is meant for longer term keeping.  Backup storage is often not on the same device as the TOE and may be physically hundreds of miles away.  In practice, the TOE encrypts the keys using an approved method before sending them to backup storage.  TOEs may perform cryptographic key escrow in which it entrusts a third party with access to the private or secret keys.  In practice, the TOE protects the keys using a cryptographic key access method agreed upon with the escrow agent before sending them._
```
smuellerDD: On escrow: Should the escrow mechanism be linked to some objective for the environment or OSP considering that the TOE has no means to enforce this?

cPP_DSC.adoc
```diff
@@ -1023,6 +1023,94 @@ FCS_CKM.2.1:: The TSF shall distribute cryptographic keys in accordance with a s
...
++
+*_Cryptographic key destruction* - The attributes of the cryptographic key should indicate the necessary conditions needed to destroy it.  Often this will be an authentication indicating authorization to destroy it.  This SFR would apply to keys stored in the TOE, and FCS_CKM.6 would apply for destruction.  Otherwise, the destruction of keys stored outside the TOE is not enforceable by this SFR._
++
+*_Key Access Methods_*
++
+_This SFR lists several methods for protecting the confidentiality of keys prior to authorizing their use, archival, backup, escrow, or recovery.  This protection is often afforded only to the private or secret portion of cryptographic keys.  Some methods offer integrity protection as well.  These methods may not apply to modification of attributes especially if the attributes do not need confidentiality.  These methods may not apply to destruction since a TOE can only destroy keys within its boundaries, and it is premused these keys are already in the boundary._
```
smuellerDD: On key destruction outside the TOE boundary: Should we link that to some objective of the environment or OSP?

cPP_DSC.adoc
```diff
@@ -1023,6 +1023,94 @@ FCS_CKM.2.1:: The TSF shall distribute cryptographic keys in accordance with a s
...
++
+*_Key encapsulation* - If the STF claims key encapsulation, then it should use approved methods such as those listed in CKM.2._
```
smuellerDD: editorial: ... STF ... -> SFR?

smuellerDD: editorial: ... should use ... -> must use?

cPP_DSC.adoc
```diff
@@ -1023,6 +1023,94 @@ FCS_CKM.2.1:: The TSF shall distribute cryptographic keys in accordance with a s
...
++
+*_Key wrapping* - If the STF claims key wrapping, then it should use approved methods such as KW, or KWP._
```
smuellerDD: STF -> SFR

smuellerDD: again: should -> must?

cPP_DSC.adoc
```diff
@@ -1023,6 +1023,94 @@ FCS_CKM.2.1:: The TSF shall distribute cryptographic keys in accordance with a s
...
++
+*_Key encryption* - If the STF claims key encryption, then it should use approved methods such as those in FCS_COP.1/SK._
```
smuellerDD: STF-> SFR

smuellerDD: again: should -> must?

cPP_DSC.adoc
```diff
@@ -1023,6 +1023,94 @@ FCS_CKM.2.1:: The TSF shall distribute cryptographic keys in accordance with a s
...
++
+*_Authentication for Key Access Methods_*
++
+_The PP/ST should contain FIA SFRs for authenticating users.  See NIST SP 800-63B “Digital Identity Guidelines” for guidance on methods for authenticating users for access to keys for any purpose._
```
smuellerDD: should -> must?

woodbe:
@smuellerDD all the text above comes straight from the crypto catalog. I do think we should discuss the specific list of "what you can do" and the accompanying descriptions. I actually think we may want to restrict the options available from the list in the catalog, but I also suspect that the catalog text may change some (as you note, several of these seem to be language issues or a lack of clarity).

@jdonndelinger do you have an update to the catalog we could look at that may resolve some of these?

woodbe:
Merged based on 10/12 call

PR #167, opened 2023-10-10
update storage description

### LINKED ISSUE #32: CCDB AU/002 - Section 2.2.2 Services
Table 1 – row for Protect

### Comment:
The descriptive text for the Protect row seems to indicate that the Protect service helps with SDOs outside the TOE boundary whereas Figure 4 on the previous page indicates that is handled by the Propagate function.

### Proposal:
Consider Protect text like "… SDOs inside the boundary."

### Priority: 1

woodbe:
This is to close #32. The text has been updated to emphasize local storage vs transmission (for propagate).

### DIFF
--- a/cPP_DSC.adoc
+++ b/cPP_DSC.adoc
@@ -193,7 +193,7 @@ DSCs provide seven core security services to a platform as illustrated in <<Core
 |The DSC shall create SDOs from parsed or generated SDEs and attributes using binding mechanisms to apply integrity protection to the SDEs together with their attributes.
 
 |Protect
-|The DSC shall manage protected storage for all SDOs. Different DSCs may implement local storage internal to the DSC boundary or utlilize external storage outside the DSC boundary. A DSC shall maintain the integrity and confidentiality (if required) of SDOs both inside the boundary and stored outside the boundary.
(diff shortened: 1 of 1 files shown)


robertmclemons (commented):
I would remove "Different" since I assume the same DSC could use both internal and external storage.

cPP_DSC.adoc
```diff
@@ -193,7 +193,7 @@ DSCs provide seven core security services to a platform as illustrated in <<Core
 |The DSC shall create SDOs from parsed or generated SDEs and attributes using binding mechanisms to apply integrity protection to the SDEs together with their attributes.
 
 |Protect
-|The DSC shall manage protected storage for all SDOs. Platform users (through client applications) can store SDOs either inside or outside the DSC boundary. A DSC shall maintain the integrity and confidentiality (if required) of SDOs both inside the boundary and stored outside the boundary.
+|The DSC shall manage protected storage for all SDOs. Different DSCs may implement local storage internal to the DSC boundary or utlilize external storage outside the DSC boundary. A DSC shall maintain the integrity and confidentiality (if required) of SDOs both inside the boundary and stored outside the boundary.
```
yiatsec, line 196: I'd suggest to replace "SDOs both inside the boundary and stored outside the boundary" with "SDOs stored both inside and outside the boundary."

woodbe:
I have made both edits.

yiatsec (approved):

PR #207, opened 2024-01-16
Hammering Update

### LINKED ISSUE #124: PBKDF from FDE proposal

woodbe:
This is to close #124 based on the latest discussion. This incorporates the FCS_CKM_EXT.8 from the catalog as well as the new SFR as this should completely cover the DSC0001 TD.

### DIFF
--- a/cPP_DSC.adoc
+++ b/cPP_DSC.adoc
@@ -3465,26 +3465,30 @@ This family defines requirements for key life cycle operations.
 [#img-FCS_CKM_EXT] 
 [ditaa,"FCS_CKM_EXT"]
 ....
-                           
                                                        +---+
-    +--------------------------------------------+  |->| 3 |
+                                                    +->| 3 |
+                                                    |  +---+
+    +--------------------------------------------+  |
     |                                            |  |  +---+
-    | FCS_STG_EXT Cryptographic Key Storage      +--+
+    | FCS_STG_EXT Cryptographic Key Storage      +--+->| 7 |
     |                                            |  |  +---+
-    +--------------------------------------------+  |->| 7 |
-                                                       +---+    
-                                                    
+    +--------------------------------------------+  |
+                                                    |  +---+
+                                                    +->| 8 |
+                                                       +---+
 ....
 
 FCS_CKM_EXT.3 The cryptographic key access applies primarily to the storage of keys for future use and retrieval of keys for immediate use by the TOE. There may be some overlap in primitives used in other SFRs, but the end goals here are to protect the confidentiality and authenticity of the keys while in storage.
 
 FCS_CKM_EXT.7 This SFR captures methods for key agreement in which multiple parties contribute material used to derive the shared key used by each party to encrypt and decrypt messages to and from each other. The derived key can be used either as a symmetric key, keyed-hash key, or cryptographic key for key derivation functions.
 
-*Management: FCS_CKM_EXT.3, FCS_CKM_EXT.7*
+FCS_CKM_EXT.8 Password key derivation is different from regular key derivation since for key derivation it is expected that the input parameters have full entropy compared to the expected key strength and the passwords for password-based key derivation have limited entropy.  One must add additional constraints, work, or entropy to increase the security of password-based key derivation algorithms that suffer from a lack of entropy induced by passwords sourced by human memory.  These password-based key derivation algorithms should result in derived keys that have sufficient security.
+
+*Management: FCS_CKM_EXT.3, FCS_CKM_EXT.7, FCS_CKM_EXT.8*
 
 No specific management functions are identified.
 
-*Audit: FCS_CKM_EXT.3, FCS_CKM_EXT.7*
+*Audit: FCS_CKM_EXT.3, FCS_CKM_EXT.7, FCS_CKM_EXT.8*
 
 There are no auditable events foreseen.
 
@@ -3492,10 +3496,8 @@ There are no auditable events foreseen.
 [horizontal]
 Hierarchical to:: No other components.
 
-Dependencies:: [FDP_ITC.1 Import of user data without security attributes, +
-or +
-FDP_ITC.2 Import of user data with security attributes, +
-or +
+Dependencies:: [FDP_ITC.1 Import of user data without security attributes, or +
+FDP_ITC.2 Import of user data with security attributes, or +
 FCS_CKM.1 Cryptographic key generation, or +
 FCS_CKM.5 Cryptographic key derivation, or +
 FCS_CKM_EXT.8 Password-based key derivation] +
@@ -3511,10 +3513,8 @@ FCS_CKM_EXT.3.1:: The TSF shall use specified cryptographic key access methods [
 [horizontal]
 Hierarchical to:: No other components.
 
-Dependencies:: [FDP_ITC.1 Import of user data without security attributes, +
-or +
-FDP_ITC.2 Import of user data with security attributes, +
-or +
+Dependencies:: [FDP_ITC.1 Import of user data without security attributes, or +
+FDP_ITC.2 Import of user data with security attributes, or +
 FCS_CKM.1 Cryptographic key generation, or +
 FCS_CKM.5 Cryptographic key derivation, or +
 FCS_CKM_EXT.8 Password-based key derivation] +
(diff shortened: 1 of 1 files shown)


woodbe:
On going through other issues, it occurs to me that we seem to have a limit specified in terms of how many attempts can be performed within a time period. In FIA_SOS.2.1 it is specified that for any single attempt, it should be at least 1:1M that a random attempt would succeed, and that within a one minute period the change should be no more than 1:100K. This would seem to imply that the limit for attempts should be no more than 10 attempts in a 1 minute period. This is for generated data, but why would the rate of attempts be different if say an x509 cert was passed to the DSC to use for authentication?

It would seem like this is what we should specify in the FIA_HMR_EXT.1 SFR as the rate, but I want to point this out because it is coming from FIA_SOS.1.

cPP_DSC.adoc
```diff
@@ -1941,6 +1943,14 @@ _If [.underline]#prevent all future authorization attempts indefinitely (i.e., l
 +
 _If [.underline]#factory reset the TOE wiping out all non-permanent SDOs, as described by FDP_FRS_EXT.2# is selected in FIA_AFL_EXT.1.3, the selection-based SFR FDP_FRS_EXT.2 must be claimed._
 
+==== FIA_HMR_EXT.1 Online Authentication Hammering Prevention
+
+FIA_HMR_EXT.1 Online Authentication Hammering Prevention
+
+FIA_HMR_EXT.1.1:: The TSF shall limit the number of authentication attempts for each counter specified in FIA_AFL_EXT.1.1 by [.underline]#[selection: using _[assgnment: integer number greater than or equal to 1,000 iterations]_ of PBKDF as specified in FCS_CKM_EXT.8.1, specifying a maximum of _[assignment: number of attempts]_ that can occur duing _[assignment: time period]_, setting the minimum time period between consecutive attempts to _[assignment: time period]_, preventing future attempts for a time period as defined in FIA_AFL_EXT.1.3, _[assignment: other anti-hammering method]_]#.
```
smuellerDD, line 1950: Once again, this will not work with Apple: they have a much smaller round count of PBKDF but "back it up" with another cryptographic operation that is a similar iteration operation as PBKDF.

woodbe: So I wrote this intentionally this way, but left an open assignment at the end. So basically if someone just used PBKDF, they should have a high iteration count, or they can do other things related to time periods, etc. At the end though, if they are doing something not explicitly mentioned here, they can describe that in the open assignment.

My understanding of the way that Apple works (admittedly not very specific) is that they could possibly meet the second selection (basically specifying they are limiting the number of attempts that can be performed over time, with the method for doing so being their other operation), but they could definitely describe it in the open assignment.

This could be further refined but I don't want to make something specific for Apple, but to make sure that Apple can either describe what they are doing for the claim or have something generic enough they could use as their selection and then provide their explanation to meet it.

smuellerDD:
Am Mittwoch, 17. Januar 2024, 15:50:25 CET schrieb Brian Wood:

Hi Brian,

> @woodbe commented on this pull request.
> 
> > @@ -1941,6 +1943,14 @@ _If [.underline]#prevent all future authorization
> > attempts indefinitely (i.e., l
>  +
>  _If [.underline]#factory reset the TOE wiping out all non-permanent SDOs,
> as described by FDP_FRS_EXT.2# is selected in FIA_AFL_EXT.1.3, the
> selection-based SFR FDP_FRS_EXT.2 must be claimed._
> 
> +==== FIA_HMR_EXT.1 Online Authentication Hammering Prevention
> +
> +FIA_HMR_EXT.1 Online Authentication Hammering Prevention
> +
> +FIA_HMR_EXT.1.1:: The TSF shall limit the number of authentication attempts
> for each counter specified in FIA_AFL_EXT.1.1 by [.underline]#[selection:
> using _[assgnment: integer number greater than or equal to 1,000
> iterations]_ of PBKDF as specified in FCS_CKM_EXT.8.1, specifying a maximum
> of _[assignment: number of attempts]_ that can occur duing _[assignment:
> time period]_, setting the minimum time period between consecutive attempts
> to _[assignment: time period]_, preventing future attempts for a time
> period as defined in FIA_AFL_EXT.1.3, _[assignment: other anti-hammering
> method]_]#.
> 
> So I wrote this intentionally this way, but left an open assignment at the
> end. So basically if someone just used PBKDF, they should have a high
> iteration count, or they can do other things related to time periods, etc.
> At the end though, if they are doing something not explicitly mentioned
> here, they can describe that in the open assignment.
> 
> My understanding of the way that Apple works (admittedly not very specific)
> is that they could possibly meet the second selection (basically specifying
> they are limiting the number of attempts that can be performed over time,
> with the method for doing so being their other operation), but they could
> definitely describe it in the open assignment.
> 
> This could be further refined but I don't want to make something specific
> for Apple, but to make sure that Apple can either describe what they are
> doing for the claim or have something generic enough they could use as
> their selection and then provide their explanation to meet it.

Whoops, now I seem to have misread the formatting: the 1000 iteration count is 
not mandatory but a selection. With that, please disregard my comment: change 
is approved from my side.


Ciao
Stephan

yiatsec:
I don't think we need to add the new SFR, FIA_HMR_EXT.1.1, to address the PBDKF issue. The current selection-based  PBKDF-related SFR, B.1.8. FCS_COP.1/PBKDF, is taken from the Crypto Catalog by the CCDB Crypto WG and it covers all options we need.

The Application Note 54 may need an update because the referenced "PBK row", "KeyDrv4", "KeyDrv5", "KeyDrv6" does not seem to exist in the working version of DSC cPP .adoc file.

yiatsec:
I don't think we need to add the new SFR, FIA_HMR_EXT.1.1, to address the PBDKF issue. The current selection-based  PBKDF-related SFR, B.1.8. FCS_COP.1/PBKDF, is taken from the Crypto Catalog by the CCDB Crypto WG and it covers all options we need.

The Application Note 54 may need an update because the referenced "PBK row", "KeyDrv4", "KeyDrv5", "KeyDrv6" does not seem to exist in the working version of DSC cPP .adoc file.

woodbe:
I reopened it for discussion on the call. I am not sure I concur with the call to close this, but if everyone agrees that the change to add FCS_CKM_EXT.8 is sufficient to meet the DSC0001 TD I'm OK, but v2 will no longer have the explicit DSC0001 interpretation, so it has to be clear that we aren't backing that change out in some manner to lead to needing an immediate TD after publication.

woodbe:
So after all the discussions, what I think may be simplest, is that we ignore the idea of anti-hammering at this point and just change the FCS_COP.1/PBKDF to FCS_CKM_EXT.8 from the catalog. The new SFR doesn't have any PBKDF rounds requirement, so it is sufficient to resolve the concern for the original TD, and it doesn't seem like there is agreement on the need for actual online anti-hammering itself, even though that was sort of wrapped up in the original SFR.

I am proposing that this pull request be modified (or pulled and resubmitted) to only change those two SFRs and that we create a new issue tagged for v3 to reconsider anti-hammering as a separate discussion later.

slpotte:
I concur with Brian's proposal.

woodbe:
I have removed the anti-hammering portion of the pull request so it is now what had been discussed on the call. Please approve for merging.

cPP_DSC.adoc
```diff
@@ -3064,6 +3064,16 @@ _If deriving a secret IV, select [.underline]#KDF-SHA, KDF-MAC-1S, or KDF-MAC-2S
...
+
+FCS_CKM_EXT.8.1:: The TSF shall perform password-based key derivation functions in accordance with a specified cryptographic algorithm [_HMAC-[selection: SHA-256, SHA-384, SHA-512]_], with iteration count of _[assignment: number of iterations]_ using a randomly generated salt of length _[assignment: length of salt]_ and output cryptographic key sizes _[selection: 128, 192, 256]_ bits that meet the following standard: [NIST SP 800-132].
+
+_Application Note {counter:remark_count}_:: _The TSF must condition a password into a string of bits prior to using it as input to algorithms that form SKs and KEKs. The TSF can perform conditioning using one of the identified hash functions or the process described in NIST SP 800-132; the ST author selects the method used. NIST SP 800-132 requires the use of a pseudo-random function (PRF) consisting of HMAC with an approved hash function._
++
+_The TOE must claim this requirement if it claims FCS_CKM.1/SKG and selects an algorithm in the PBK row or claims FCS_CKM.5 and selects at least one of [.underline]#KeyDrv4#, [.underline]#KeyDrv5#, or [.underline]#KeyDrv6# AND uses password-based key derivation to create at least one of the inputs._
```
yiatsec, line 3075: Please remove "The TOE must claim this requirement if it claims FCS_CKM.1/SKG and selects an algorithm in the PBK row or claims FCS_CKM.5 and selects at least one of [.underline]#KeyDrv4#, [.underline]#KeyDrv5#, or [.underline]#KeyDrv6# AND uses password-based key derivation to create at least one of the inputs." The referenced "PBK row", "KeyDrv4", "KeyDrv5", and "KeyDrv6" no longer exist.

yiatsec (changes requested):
Please remove "The TOE must claim this requirement if it claims FCS_CKM.1/SKG and selects an algorithm in the PBK row or claims FCS_CKM.5 and selects at least one of [.underline]#KeyDrv4#, [.underline]#KeyDrv5#, or [.underline]#KeyDrv6# AND uses password-based key derivation to create at least one of the inputs." The referenced "PBK row", "KeyDrv4", "KeyDrv5", and "KeyDrv6" no longer exist.

woodbe:
Closing as I couldn't quite get my head around the merge conflicts due to the one change that moved the arch table to the end.

PR #208, opened 2024-01-16
sfr arch table move

### LINKED ISSUE #185: Table in 6.1
Shall really the here listed SFRs always be completely fulfilled? It hast he appearance and seems to be very complex. Is this the intention?

woodbe:
This is to close #185 by moving the table into the appendices instead of in the middle of the SFRs.


smuellerDD (approved):
The diff does not allow seeing that the table is just moved. But if it is just moved, I see no issue.

woodbe:
Well that is all I did, but yeah, it is very difficult to see. I think it gets really confused because of the very large block being moved. This is basically pulling the section (I think it was 6.1) and making it into a new Annex section at the end (without any text changes).

PR #222, opened 2024-01-23
editing of roles

woodbe:
Attempt to better write the roles. Still needs more review, but should be most of the way there.

### DIFF
--- a/cPP_DSC.adoc
+++ b/cPP_DSC.adoc
@@ -325,11 +325,11 @@ As the DSC is generally a component within a larger system or platform, the role
 The ADM-R role provides sufficient privileges to manage the functionality of the DSC. As a role designed for the administrator, this role may be responsible for the following:
 
 * Manage access control for SDOs (does not mean the contents of any particular SDO can be read by the administrator)
-* Manage the initial configuration of the DSC
+* Manage the configuration of the DSC
 
 The MFGADM-R role, if explicitly defined in a DSC, may include the following responsibilities (which may be part of the ADM-R role otherwise):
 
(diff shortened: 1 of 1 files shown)


cPP_DSC.adoc
```diff
@@ -312,21 +312,42 @@ The following paragraphs describe the relationships between elements of the DSC.
...
+As the DSC is generally a component within a larger system or platform, the roles of the DSC are specific to the DSC. While there may be matching roles between the DSC and its platform, the roles here are specifically those in the DSC and are independent of any defined on the platform.
+
+The ADM-R role provides sufficient privileges to manage the functionality of the DSC. As a role designed for the administrator, this role may be responsible for the following:
+
+* Manage access control for SDOs (does not mean the contents of any particular SDO can be read by the administrator)
+* Manage the initial configuration of the DSC
+
+The MFGADM-R role, if explicitly defined in a DSC, may include the following responsibilities (which may be part of the ADM-R role otherwise):
+
+* Manage the pre-installed SDOs and initial configuration of the DSC
```
smuellerDD, line 332: why only initial configuration? Shouldn't we say "... and runtime configuration, if applicable ..."?

smuellerDD (approved):
apart from the small issue, the changes look good

cPP_DSC.adoc
woodbe: The way I had interpreted the descriptions initially seemed to focus more on the initial configuration, so I don't know. I don't think it matters too much and have no problem adding that, but I don't know if we really are expecting the manufacturer to control the runtime configuration. Easy to add though if we want it.

PR #234, opened 2024-02-01
PBKDF fix

woodbe:
#207 had some issues because of an earlier pull request that had me scratching my head to resolve the conflicts. This is the same but based on the latest working branch updates.

Please approve.

### DIFF
--- a/cPP_DSC.adoc
+++ b/cPP_DSC.adoc
@@ -2028,7 +2028,7 @@ The following rationale provides justification for each security objective for t
 |FPT_STM.1 
 |This requirement provides reliable system time services that may be used to determine when excessive authentication failure attempts have been made.
 
-| FIA_AFL_EXT.2 (selection-based)
+|FIA_AFL_EXT.2 (selection-based)
 |This requirement defines how access to an SDO is restored if excessive authentication failures trigger a lock on it.
 
 .10+|O.AUTHORIZATION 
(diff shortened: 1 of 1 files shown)


woodbe:
@yiatsec, you said:

> Please remove "The TOE must claim this requirement if it claims FCS_CKM.1/SKG and selects an algorithm in the PBK row or claims FCS_CKM.5 and selects at least one of [.underline]#KeyDrv4#, [.underline]#KeyDrv5#, or [.underline]#KeyDrv6# AND uses password-based key derivation to create at least one of the inputs." The referenced "PBK row", "KeyDrv4", "KeyDrv5", and "KeyDrv6" no longer exist.

In Table 17 for FCS_CKM.5.1, it looks like the same lines exist, they are different though, and I really can't tell you how the old and new tables actually map to each other.

I'm happy to make a change, but I suspect it should be an edit pointing to the FCS_CKM.5 table and not removing that line completely. But that said, what rows are correct?

yiatsec:
The "List of Standards" column in the FCS_CKM.5 table does not have NIST SP 800-132 (the NIST PBKDF standard) on any row. So, there are no rows to be pointed to. The old and new tables for FCS_CKM.5 do not have a one-to-one mapping. The new table focuses only on Key-Based KDF and leaves out Password-based KDF, which is a cleaner approach. The PBDKF is covered and only covered by FCS_CKM.8.

Therefore, removing that entire paragraph is safe to do.

yiatsec (approved):

PR #238, opened 2024-02-09
Base crypto update to SD

woodbe:
This is a base update to the SD that updates the SFRs to match the current cPP and moving some of the requirements around as they have moved to match the cPP.

There are several markers of [IMPORTANT] in the document that point out where the requirements have either been added or changed (such as FCS_CKM.2 changed from establishment to distribution). These would seem to be the primary areas of focus initially since they are as yet undefined.

### DIFF
--- a/SD_DSC.adoc
+++ b/SD_DSC.adoc
@@ -248,7 +248,7 @@ For definitions of standard CC terminology, see [CC] part 1.
 |AES 
 |Advanced Encryption Standard
 
-|CA 
+|CApp
 |Client Application
 
 |CBC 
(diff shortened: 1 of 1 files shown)


woodbe:
I will be basing future pull requests on this one.

slpotte:
In new line 554, there is a reference to FCS_COP.1/KAT.  However, it appears to have been removed; see old line 684-797.

woodbe:
Yeah, I actually see a few more of those. This initial change was mainly about getting all the headers correct and the rough content moved where needed. I haven't reviewed the individual content yet, but that will come in further updates. There are probably a lot more, but need to start somewhere.

slpotte (approved):
Per Brian, will fix outdated references to /KAT later.

Issue #239, opened 2024-02-12
Add phrase to FCS_CKM_EXT.7.1 App Note 5


JDavid-Thompson:
In Para 3, suggest adding "and there are two parties" to the parenthetical so it reads "(e.g., if each share is 128 bits and there are two parties, then the result after concatenation is a 256-bit key, which is appropriate only for AES-256-CMAC)"

In Para 4, suggest changing “For the KDF functions and XOR, each party may have to use an asymmetric method from FCS_CKM_EXT.7 to transmit their shares to each other” to “For the KDF functions and XOR, each party must transmit their shares securely, either by encrypting with an asymmetric method from FCS_CKM_EXT.7 or by procedural means, such as sending them by courier.”

JDavid-Thompson:
Brian,

I think I made a pull request for #239.  I’m still a little unsure of all the github terms and operators.

-Dave

woodbe:
I saw it

woodbe:
@woodbe will review the current crypto catalog and provide a response here (or a PR)

woodbe:
The second proposed change has been already updated (differently) in the catalog, and that text actually makes more sense.

PR #243, opened 2024-02-15
App Note for FCS_CKM_EXT.7

JDavid-Thompson:
Change the wording in two places in the App Note for FCS_CKM_EXT.7.

### DIFF
--- a/cPP_DSC.adoc
+++ b/cPP_DSC.adoc
@@ -799,9 +799,9 @@ _Application Note {counter:remark_count}_:: _This SFR captures methods for multi
 +
 _FCS_CKM.5 defines KDF-CTR, KDF-FB, and KDF-DPI._
 +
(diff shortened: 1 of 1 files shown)


cPP_DSC.adoc
```diff
@@ -799,9 +799,9 @@ _Application Note {counter:remark_count}_:: _This SFR captures methods for multi
...
 +
-_For the KDF functions and XOR, each party may have to use an asymmetric method from FCS_CKM_EXT.7 to transmit their shares to each other. Key shares may also come from a token, in which case, TOEs may use key access methods in FCS_CKM.3 to authorize access and use of those keys in this SFR._
+_For the KDF functions and XOR, each party must transmit their shares securely, either by encrypting with an asymmetric method from FCS_CKM_EXT.7 or by procedural means, such as sending them by courier. Key shares may also come from a token, in which case, TOEs may use key access methods in FCS_CKM.3 to authorize access and use of those keys in this SFR._
```
woodbe, line 804: I think that instead of "sending them by courier" it should just be "out of band" and not be an example. That method is basically out of scope here, just that they are there.

woodbe (changes requested):

woodbe:
This is to close #239

cPP_DSC.adoc
slpotte: This is worded as a requirement.  Can we point to an existing SFR that covers this?  E.g. FCS_CKM.2 Key Distribution.

woodbe: That would work for me, that at least keeps it as part of the TOE.

woodbe:
This is considered as resolved by #340 as the group does not think that we need to include out-of-band methods of key exchange for a DSC.

PR #250, opened 2024-02-27
Sd fcs ckm updates

woodbe:
This has changes to FCS_CKM in section 2.1.1 in the SD. I didn't make any changes for FCS_CKM.6 since #229 should have further changes.

FCS_CKM_EXT.7 looks like we may end up needing to add a whole set of algorithm tests here?


slpotte (commented):
For ECDH, you'll need FCS_CKM.1/AKG to generate private keys and FCS_CKM_EXT.7 for the key establishment algorithm.  The crypto working group is changing the rows for ECDSA to just ECC to make sure that the private key generation applies to both ECDH and ECDSA.

Issue #235, opened 2024-02-01
Not to record key value before Key Destruction (FCS_CKM.4)


yiatsec:
In DSC cPP SD v1.0, the test requirement for FCS_CKM.4 Cryptographic Key Destruction includes a step to record the value of the key or keyring material, which will be used to compare against the value after the key destruction. Due to security concerns, some TOEs will never allow output plaintext keys/key materials. Therefore, the test requirement should be relaxed to accommodate other means (e.g., compare the hash value) to verify the destruction of the original key/key material.

PR #251, opened 2024-03-05
Change Test FCS_CKM.6

yiatsec:
Change to address issue https://github.com/DSC-iTC/cPP/issues/235

### DIFF
--- a/SD_DSC.adoc
+++ b/SD_DSC.adoc
@@ -484,38 +484,16 @@ The following tests require the developer to provide access to a test platform t
 
 The evaluator shall perform the following tests:
 
-*Test 1:* Applied to each key or keying material held as plaintext in volatile memory and subject to destruction by overwrite by the TOE (whether or not the plaintext value is subsequently encrypted for storage in volatile or non-volatile memory).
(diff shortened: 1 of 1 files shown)


woodbe:
Discussion results:

- Leave current test 1, but make conditional to implementations where keys are directly accessible
- take proposed test 1 and make conditional to implementations where keys are referenced
- Force choice of one of the test types (grouping test 1 and test 2 together for the selection)

Merge at the moment while being aware that updated requirements may come from @JDavid-Thompson and the Crypto WG

JDavid-Thompson:
I don’t know what the objections to my version of FCS_CKM.6 are, other asserting that it is not necessary.

  *   Do you not agree with my interpretation of FDP_RIP.1?
  *   Do you agree with that, but think that wear-leveled flash meets that interpretation?

-Dave

JDavid-Thompson:
Brian,

OK.  Thanks for the clarification.

-Dave

jvdsn:
I don't believe the existing tests should be changed this much to address #235. I opened a new PR #292 with smaller changes.

PR #254, opened 2024-04-09
Section 2.3 and 2.4 changes

jvdsn:
Changes justified inline.

PR #256, opened 2024-04-10
user/roles update

### LINKED ISSUE #33: CCDB AU/003 - Section 2.2.4 DSC Characteristics
### Comment:
This section introduces the “user” concept but does not really use it to clarify DSC function.  For example: FDP_ACC just uses Subjects, Objects and Operations.  Application Note 32 indicates that user is a generic term that can be admin or client application roles.  Section 2.4.4 talks about roles and key hierarchies.  The concept of roles seems more precise when explaining the uses of the DSC. 

### Proposal:
In the context of a DSC is might be better to concentrate on the concept of roles rather than users to explain functions.  The text in Section 2.2.4 could perhaps mention roles to explain how the administrator role sets up key information so that client applications can use the DSC to prove various assertions.  This fits in with the use-case for these devices where the DSC is not directly controlled by the device end-user.

### Priority: 1

woodbe:
This is to close #33 by trying to update the users definition.

I removed Users from the in question section and made it a lower subheading and then tweaked the text. I also added roles to the list but then left that light and pointed to the roles section. I also moved the roles section to this one from the Key Hierarchy as it makes more sense there.

### DIFF
--- a/cPP_DSC.adoc
+++ b/cPP_DSC.adoc
@@ -230,7 +230,7 @@ Subjects: The following list contains the fundamental actors in the expected ope
 
 * S.DSC - DSC with security attribute DSC.ID, which is the identity of the DSC
 * S.Admin - Admin (an authorized administrator with special privileges) security attribute - See <<Roles>> for more discussion on administrator roles.
-* S.CApp - CApp (i.e. an authorized user or an application with a verifiable identity) with security attribute CApp.ID - See <<Roles>> for more discussion on user roles.
+* S.CApp - Client Application (CApp) (i.e. an authorized user or an application with a verifiable identity) with security attribute CApp.ID - See <<Roles>> for more discussion on user roles.
 * S.EPS - External Platform Storage (EPS) (e.g. transient SDE/SDO source and destination, in the case of data imported and exported for the sole use inside the DSC). In the case of a passive EPS, the DSC will properly protect the integrity and confidentiality of the objects it stores and retrieves from there. In the case of an active EPS with security attribute EPS.ID, the DSC and EPS may choose to create a secure channel through which they will pass objects back and forth.
 
 Roles: Users of the DSC are assigned to Roles which enumerate the permissions which are granted to the Objects and Operations. See <<Roles>> for more discussion on roles.
--- a/cPP_DSC.adoc
+++ b/cPP_DSC.adoc
@@ -220,13 +220,11 @@ This collaborative Protection Profile (cPP) assumes a DSC will contain a Root of
 
 The security functional requirements rely on the following characteristics of the DSC:
 
-* Users
 * Subjects
+* Roles
 * Objects
 * Security Attributes
 * Operations
-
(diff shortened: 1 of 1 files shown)


cPP_DSC.adoc
```diff
@@ -220,21 +220,21 @@ This collaborative Protection Profile (cPP) assumes a DSC will contain a Root of
...
 
 * S.DSC - DSC with security attribute DSC.ID, which is the identity of the DSC
 * S.Admin - Admin (an authorized administrator with special privileges) security attribute - See <<Roles>> for more discussion on administrator roles.
 * S.CApp - CApp (i.e. an authorized user or an application with a verifiable identity) with security attribute CApp.ID - See <<Roles>> for more discussion on user roles.
 * S.EPS - External Platform Storage (EPS) (e.g. transient SDE/SDO source and destination, in the case of data imported and exported for the sole use inside the DSC). In the case of a passive EPS, the DSC will properly protect the integrity and confidentiality of the objects it stores and retrieves from there. In the case of an active EPS with security attribute EPS.ID, the DSC and EPS may choose to create a secure channel through which they will pass objects back and forth.
 
+Roles: Users of thge DSC are assigned to Roles which enumerate the permissions which are granted to the Objects and Operations. See <<Roles>> for more discussion on roles.
```
jvdsn, line 236: typo thge

cPP_DSC.adoc
```diff
@@ -220,21 +220,21 @@ This collaborative Protection Profile (cPP) assumes a DSC will contain a Root of
...
-
 Subjects: The following list contains the fundamental actors in the expected operational use cases of the DSC. The first three are active actors, while the fourth is usually passive but could be active.
 
 * S.DSC - DSC with security attribute DSC.ID, which is the identity of the DSC
 * S.Admin - Admin (an authorized administrator with special privileges) security attribute - See <<Roles>> for more discussion on administrator roles.
 * S.CApp - CApp (i.e. an authorized user or an application with a verifiable identity) with security attribute CApp.ID - See <<Roles>> for more discussion on user roles.
```
jvdsn, line 233: Probably define CApp here ("Client Application") since its definition above was removed.

cPP_DSC.adoc
```diff
@@ -271,44 +271,12 @@ Operations: The following list contains the expected operations of a DSC.
 * OP.Export (See Propagate) - The DSC may export SDOs, SDEs, Authorization Data, Platform State, or Anti-Replay Tokens to a CApp or EPS with the proper authorization from the owner of each object. In the case of EPS, the DSC will bind the objects to the DSC in such a way as to deny other DSCs or entities the ability to import, use, modify, attest, store, export, or destroy them. The DSC may export Session Contexts only to an EPS binding it in the same way as above.
 * OP.Destroy (See Purge) - The DSC may purge SDOs, SDEs, Authorization Data, Platform State, Anti-Replay Tokens, or Session Contexts in protected storage with proper authorization from the owner of each object.
 
-=== TOE Use Cases
```
jvdsn, line 274: This change will conflict with #254

cPP_DSC.adoc
woodbe: Technically it is defined in the glossary, just in this document, for some reason, that is at the end in Annex G and H (for acronyms). We shouldn't need to define things that are already in the glossary, though we may want to consider moving it from the back to the front.

cPP_DSC.adoc
woodbe: Technically it is defined in the glossary, just in this document, for some reason, that is at the end in Annex G and H (for acronyms). We shouldn't need to define things that are already in the glossary, though we may want to consider moving it from the back to the front.

jvdsn: Currently, the document quite often defines abbreviations even if they are defined in the glossary or list of abbreviations. It's just a suggestion to define it here.

woodbe: This has been added where CApp is first used.

jvdsn (approved):

PR #266, opened 2024-04-23
SD updates for FDP (mandatory)

woodbe:
This is a set of changes for the SD in the FDP requirements. Nothing major, mostly editorial.

### DIFF
--- a/SD_DSC.adoc
+++ b/SD_DSC.adoc
@@ -1437,7 +1437,7 @@ There are no AGD evaluation activities for this component.
 
 If the TOE stores SDEs and authorization data inside the TSF, the evaluator shall ensure that external interfaces cannot extract this data in plaintext. 
 
-In this case, use the evaluation activities of the FPT_PHP.3 if [.underline]#protected storage# is selected, FCS_COP.1/SKC if [.underline]#symmetric encryption using…# is selected, and FCS_COP.1/KAT if [.underline]#key wrapping using…# is selected.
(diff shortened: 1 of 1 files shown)


SD_DSC.adoc
```diff
@@ -1451,19 +1451,19 @@ There are no KMD evaluation activities for this component.
 
 ====== TSS
 
-The evaluator shall confirm that the ST author describes the methods for protecting the integrity of SDOs stored with the TOE, and shall identify the iteration of FCS_COP.1/Hash or FCS_COP.1/HMAC that covers any cryptographic algorithm used. The evaluator shall also confirm that the TSS describes the response upon the detection of an integrity error.
+The evaluator shall confirm that the ST author describes the methods for protecting the integrity of SDOs stored with the TOE, and shall identify the iteration of FCS_COP.1/Hash, FCS_COP.1/KeyedHash, FCS_COP.1/SigVer or FCS_COP.1/KeyWrap that covers any cryptographic algorithm used. The evaluator shall also confirm that the TSS describes the response upon the detection of an integrity error.
```
jvdsn, line 1454: Should this change also be made in the cPP? Right now the cPP says `shall identify the iteration of FCS_COP.1/Hash or FCS_COP.1/KeyedHash that covers any cryptographic algorithm used`

What about FCS_COP.1/CMAC? See #268

woodbe: It wasn't there originally. I would prefer to handle that as a separate update though since it probably has to be worked into more places than just here. This is purely to correct what is there now. If we are going to add CMAC, it should be done separately.

jvdsn (approved):

PR #267, opened 2024-04-23
FDP_DSI.2 edit

woodbe:
The app note did not specify all the algorithm SFRs that are listed in the selection.

### DIFF
--- a/cPP_DSC.adoc
+++ b/cPP_DSC.adoc
@@ -1571,7 +1571,7 @@ _Application Note {counter:remark_count}_:: _This SFR deals with the mechanism t
 +
 _The cryptographic requirements for cryptographic hashes and digital signatures apply here._
 +
-_No specific requirement is placed here on the nature of the integrity protection data, but the Security Target shall describe this protection measure, and shall identify the iteration of FCS_COP.1/Hash or FCS_COP.1/KeyedHash that covers any cryptographic algorithm used._
+_No specific requirement is placed here on the nature of the integrity protection data, but the Security Target shall describe this protection measure, and shall identify the iteration of FCS_COP.1/Hash, FCS_COP.1/KeyedHash, FCS_COP.1/KeyVer or FCS_COP.1/KeyWrap that covers any cryptographic algorithm used._
 +
 _The integrity protection data in FDP_SDI.2.1 is included in the list of attributes identified in FMT_MSA.1, and protects the value of the SDEs and of the SDO security attributes._
 +


cPP_DSC.adoc
```diff
@@ -1571,7 +1571,7 @@ _Application Note {counter:remark_count}_:: _This SFR deals with the mechanism t
 +
 _The cryptographic requirements for cryptographic hashes and digital signatures apply here._
 +
-_No specific requirement is placed here on the nature of the integrity protection data, but the Security Target shall describe this protection measure, and shall identify the iteration of FCS_COP.1/Hash or FCS_COP.1/KeyedHash that covers any cryptographic algorithm used._
+_No specific requirement is placed here on the nature of the integrity protection data, but the Security Target shall describe this protection measure, and shall identify the iteration of FCS_COP.1/Hash, FCS_COP.1/KeyedHash, FCS_COP.1/KeyVer or FCS_COP.1/KeyWrap that covers any cryptographic algorithm used._
```
jvdsn, line 1574: #268 adds FCS_COP.1/CMAC but not the others. I don't have any problems with adding the others as well.

cPP_DSC.adoc
jvdsn, line 1: On line 1564, `FDP_SDI.2.1:: The TSF shall monitor *SDOs and SDEs* controlled by the TSF for [_integrity errors_] on all objects, based on the following attributes: [.underline]#*[selection: [_assignment: attribute associated with presence in protected storage_], cryptographic hash, digital signature, integrity-providing encryption algorithm as specified in FCS_COP.1/KeyWrap]*#.` should be changed to `cryptographic hash, message authentication code, digital signature, integrity-providing encryption algorithm as specified in ...`

jvdsn (changes requested):

cPP_DSC.adoc
woodbe: Is anyone actually doing this? I say that because I don't know if the reason it isn't here is because originally no one has an implementation that uses CMAC for this (it may be used for other things, I'm not sure), and so not relevant to this requirement. It doesn't matter to me, but I do not have an answer.

@slpotte any ideas?

jvdsn: @woodbe right now "cryptographic hash, digital signature, integrity-providing encryption algorithm" does not include HMAC or CMAC. So I think at least "keyed hash" should be added, but I prefer "message authentication code" since it is more general.

woodbe: Can you propose that as a separate change (it can be based on this pull request or bundled into the other one about CMAC, doesn't matter) so that at the moment we leave this as a simpler "fix what is there now" instead of "fix that and add something else we haven't talked about yet"?

jvdsn: #269

jvdsn (approved):

PR #268, opened 2024-04-24
Allow FCS_COP.1/CMAC in places where FCS_COP.1/KeyedHash is allowed

jvdsn:
Considering that the DSC PP defines FCS_COP.1/CMAC, I think we should allow it in all places where FCS_COP.1/KeyedHash is allowed.

### DIFF
--- a/cPP_DSC.adoc
+++ b/cPP_DSC.adoc
@@ -1493,7 +1493,7 @@ FDP_ITC_EXT.1 Parsing of SDEs
 
 FDP_ITC_EXT.1.1:: The TSF shall support importing SDEs using [.underline]#[selection: physically protected channels as specified in FTP_ITP_EXT.1, encrypted data buffers as specified in FTP_ITE_EXT.1, cryptographically protected data channels as specified in FTP_ITC_EXT.1]#.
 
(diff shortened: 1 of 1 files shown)


woodbe:
Let's see what the crypto WG comes back with before approving this change. It makes sense, but those requirements will be the basis for us first, so making the change now means having to see what they may do in the final version of their update.

woodbe:
Once we have #266 & #267 merged, this should include the SD changes and then it is one comprehensive update. 

Also, I suspect there may be other places this would be needed (I don't know off hand, but I suspect it is more widespread than just this one SFR). If we are going to add it, we should do all the CMAC adds in one.

woodbe:
I mean wait to see what the WG comes back with in terms of how they set those requirements up. In the meantime, check to see if there is anywhere else this should also happen sine I assume there are other places, and put that into a single pull request that covers both the cPP and the SD.

woodbe:
@jvdsn please add to the ECD and also check the SD

jvdsn:
@woodbe I checked the ECD and SD, there's no other places where this change should be made. The ECD simply says "[_assignment: method of_ _integrity verification_]"

woodbe (approved):

PR #271, opened 2024-04-25
Update FDP_ETC_EXT.2

jvdsn:
The current wording of this EA is very puzzling to me. According to Section 1.3.1 in the SD, TSF is defined as: "A set consisting of all hardware, software, and firmware of the TOE that must be relied upon for the correct enforcement of the SFRs. [CC1]"

That seems to imply to me that the TSF is part of the TOE boundary, so why is the EA talking about "data while it resides outside the TOE"? Perhaps I'm just mistaken about my interpretation of TSF.

### DIFF
--- a/SD_DSC.adoc
+++ b/SD_DSC.adoc
@@ -1325,7 +1325,7 @@ There are no KMD evaluation activities for this component.
 
 ====== TSS
 
-The evaluator shall examine the TSS to ensure that it describes how it protects the SDO references, authorization data, against access from unauthorized entities. If [.underline]#the TSF# is selected, then it should describe how it provides confidentiality of the data while it resides outside the TOE. 
+The evaluator shall examine the TSS to ensure that it describes how it protects the SDO references, authorization data, against access from unauthorized entities. In particular, it should describe how it provides confidentiality of the data while it resides outside the TOE after export.
 
 ====== AGD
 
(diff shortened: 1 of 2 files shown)


woodbe:
The SFR states who can access the data, only the TSF or authorized users. I think the description is assuming that "authorized users" is pretty clear and would be handled elsewhere, but when the TSF is selected more explicit description is required. It probably needs something that may be "if authorized users is selected, this is tested as part of XYZ".

I do not think it is correct though to change it from one to the other directly without and other change.

jvdsn:
@woodbe so TSF here includes locations outside the TOE?

woodbe:
No, the TSF is the only thing that can access the data. So if you encrypt the blob with something that only the TSF can access as opposed to encrypting is such that a user can access it. This is where the propagate terms is weird because I would normally think of that meaning that the data is sent to another system while here it is just that it is being stored someplace that the TOE does not have full control over (i.e. main storage on an Android device instead of some dedicated storage area that isn't part of the main flash).

Here "the TSF" is a user, the TOE itself. Maybe it should be TOE, and not TSF, since that makes more sense (i.e. storing it such that only the TOE can access it).

woodbe:
Change the selection from "TSF" to "TOE"

Adjust the SD to remove the "if the TSF" so it applies to both

can add that authorized users means all roles (CApps, etc)

jvdsn:
@woodbe what do you think of this suggestion?

woodbe (approved):

PR #272, opened 2024-04-25
Remove Camellia from CCMP and GCMP selections

jvdsn:
Note: these SFRs are not provided by the Crypto WG.

IEEE 802.11i says "CCMP is based on the CCM of the AES encryption algorithm". IEEE 802.11ac says "The AES algorithm is defined in FIPS PUB 197-2001. AES processing used within CCMP uses AES with either a 128-bit key (CCMP-128) or a 256-bit key (CCMP-256)."

Similarly, IEEE 802.11 also states "GCMP is based on the GCM of the AES encryption algorithm." and "The AES algorithm is defined in FIPS 197. All AES processing used within GCMP uses AES with a 128-bit key (GCMP-128) or a 256-bit key (GCMP-256)."

Why is Camellia allowed here? Is there any use case?

### DIFF
--- a/cPP_DSC.adoc
+++ b/cPP_DSC.adoc
@@ -2968,7 +2968,7 @@ _Application Note {counter:remark_count}_:: _It is acceptable for the TSF to pro
 
 FTP_CCMP_EXT.1 CCM Protocol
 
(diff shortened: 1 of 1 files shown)


woodbe (approved):

woodbe:
Please check the SD and update or comment and I will merge.

jvdsn:
@woodbe SD does not include any specific reference to Camellia for FTP_CCMP_EXT or FTP_GCMP_EXT

PR #275, opened 2024-04-25
SD FIA review (mandatory requirements)

woodbe:
Review of SD for FIA requirements that are mandatory.


SD_DSC.adoc
```diff
@@ -1485,15 +1485,15 @@ The evaluator shall also examine the TSS to determine the behaviour that will oc
 
 ====== AGD
 
-The evaluator shall examine the guidance documentation to ensure that instructions for configuring the number of successive unsuccessful authentication attempts and time period (if implemented) are provided, and that the process of unlocking the SDOs is described for each "action" specified (if that option is chosen). 
+If the administrator is able to configure any of the variables for authorization attempts, the evaluator shall examine the guidance documentation to ensure that instructions for configuring the number of successive unsuccessful authentication attempts and time period (if implemented) are provided, and that the process of unlocking the SDOs is described for each "action" specified (if that option is chosen).
```
jvdsn, line 1488: FMT_SMF.1 actually requires that the TSF is capable of performing the "Set authorization failure parameters for FIA_AFL_EXT." management function. If we want this to be optional, I believe it should be changed there.

Because this is defined in FMT_SMF.1, I feel like the first part of this sentence ("instructions for configuring the number of successive unsuccessful authentication attempts and time period (if implemented) are provided") should be evaluated as part of the AGD for FMT_SMF.1.

SD_DSC.adoc
```diff
@@ -1485,15 +1485,15 @@ The evaluator shall also examine the TSS to determine the behaviour that will oc
...
 
 ====== Test
 
 The evaluator shall perform the following tests for each method by which the TSF authorizes access the SDOs (e.g. any passwords entered as part of establishing authorization):
 
-Test 1: The evaluator shall use the operational guidance to configure the number of successive unsuccessful authorization attempts allowed by the TOE (and, if the time period selection in FIA_AFL_EXT.1.3 is included in the ST, then the evaluator shall also use the operational guidance to configure the time period after which access is re-enabled). The evaluator shall test that once the authorization attempts limit is reached, authorization attempts with valid credentials are no longer successful.
+Test 1: If the administrator is able to configure any of the variables for authorization attempts, the evaluator shall use the operational guidance to configure the number of successive unsuccessful authorization attempts allowed by the TOE. The evaluator shall test that once the authorization attempts limit is reached (whether configured by the administrator or statically specified), authorization attempts with valid credentials are no longer successful.
```
jvdsn, line 1496: Same comment as above. I believe that the act of configuring the TOE should be tested as part of FMT_SMF.1

jvdsn (changes requested):

SD_DSC.adoc
slpotte: I concur.

SD_DSC.adoc
slpotte: I concur

SD_DSC.adoc
woodbe: So we can change FMT_SMF.1, but the SFR is written to make this optional. FIA_AFL_EXT.1.2 states that the options are "static, administrator configurable variable" so by definition we are saying that in some cases the system may come configured without an option to change it. I actually think this is what is most likely for most DSC implementations anyway, but this is part of the SFR and has been, I didn't change that. The change here was because of what is in the SFR since you can't test a configurable setting if there isn't one.

woodbe:
Move FMT_SMF.1 "Set authorization failure parameters for FIA_AFL_EXT.1" to the selection for consistency.

Rest of the PR is fine.

woodbe:
I have made the change and added it to this pull request. Please approve and I will merge.

cPP_DSC.adoc
```diff
@@ -1840,12 +1840,12 @@ FMT_SMF.1 Specification of Management Functions
 
 FMT_SMF.1.1:: The TSF shall be capable of performing the following management functions: [
 +
-* _Set authorization failure parameters for FIA_AFL_EXT.1_
 * _Reset TOE to factory state for FDP_FRS_EXT.1_
 * _Configure authorization policies for TOE resources_
 +
 _[.underline]#[selection:#_
 +
+** _[.underline]#Set authorization failure parameters for FIA_AFL_EXT.1,#_
```
jvdsn, line 1848: Extremely minor comment, but I'd like if this could be un-capitalized to be consistent with the lines below.

jvdsn (approved):

PR #285, opened 2024-04-25
FIA_SOS.2 update

### LINKED ISSUE #273: Hidden throttling requirement in FIA_SOS.2 SD
https://github.com/DSC-iTC/cPP/blob/cbeef0306a44ab3af99352404baff9b6a8124f8b/SD_DSC.adoc?plain=1#L1516

In this line we have a requirement that throttling be implemented. This is not in the SFR, nor is it anywhere else in the cPP at this point (the only things we had discussed regarding throttling were all removed as they were related to the PBKDF functions).

I think this line should be removed, but am proposing it as an issue before creating the PR.

woodbe:
This is to close #273.

### DIFF
--- a/SD_DSC.adoc
+++ b/SD_DSC.adoc
@@ -1513,13 +1513,7 @@ The evaluator shall ensure that the TSS describes for each of the TSF functions 
 
 The evaluator shall ensure that the TSS states that the quality metrics provided is based on the assumption of sufficient entropy being provided in accordance with the information given in [DSC cPP] Annex D.
 
-The evaluator shall ensure that the TSS describes the restrictions implemented in order to restrict consecutive authentication attempts. (Authentication throttling)
-
-The evaluator shall ensure that the TSS describes the mechanism used to generate authorization values and documents the quality metric that the mechanism provides. The information provided in the TSS shall demonstrate that:
-
-[loweralpha]
-. The probability that a random single authentication attempt will be successful is less than one in 1,000,000; and
-. The probability that random multiple authentication attempts during a one (1) minute period will be successful is less than one in 100,000.
+The evaluator shall ensure that the TSS describes the mechanism used to generate authorization values and documents the quality metric that the mechanism provides. The information provided in the TSS shall demonstrate that the probability that a random single authentication attempt will be successful is less than one in 1,000,000.
 
 ====== AGD
 
--- a/cPP_DSC.adoc
+++ b/cPP_DSC.adoc
@@ -1622,10 +1622,7 @@ _If [.underline]#factory reset the TOE wiping out all non-permanent SDOs, as des
 
 FIA_SOS.2 TSF Generation of Secrets
 
-FIA_SOS.2.1:: The TSF shall provide a mechanism to generate *authorization data* that meet [_the following quality metrics:_
-+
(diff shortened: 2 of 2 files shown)


cPP_DSC.adoc
```diff
@@ -1622,10 +1622,7 @@ _If [.underline]#factory reset the TOE wiping out all non-permanent SDOs, as des
 
 FIA_SOS.2 TSF Generation of Secrets
 
-FIA_SOS.2.1:: The TSF shall provide a mechanism to generate *authorization data* that meet [_the following quality metrics:_
-+
-* _For each authentication attempt, the probability shall be less than one in 1,000,000 that a random attempt will be successful_
-* _For multiple attempts to authenticate during a one-minute period, the probability shall be less than one in 100,000 that a series of random attempts will be successful_].
+FIA_SOS.2.1:: The TSF shall provide a mechanism to generate *authorization data* that meet [_the metric where for each authentication attempt, the probability shall be less than one in 1,000,000 that a random attempt will be successful_].
```
jvdsn, line 1625: Looking at CC:2022, FIA_SOS.2 is really about generation of secrets. In this case, we're generating authorization data, so the secret is of course used for authentication. However, I don't think we should necessarily involve "authentication attempts" here, since I think it exceeds the scope of the SFR ever so slightly.

How about: "that meet _[the following quality metric: the probability of randomly guessing the value is less than one in 1,000,000]_"?

Of course, a random guess in this context is the same as an authentication attempt. But this wording decouples the process of authentication from the quality of the authorization data.

SD_DSC.adoc
```diff
@@ -1513,13 +1513,7 @@ The evaluator shall ensure that the TSS describes for each of the TSF functions
...
 
-The evaluator shall ensure that the TSS describes the restrictions implemented in order to restrict consecutive authentication attempts. (Authentication throttling)
-
-The evaluator shall ensure that the TSS describes the mechanism used to generate authorization values and documents the quality metric that the mechanism provides. The information provided in the TSS shall demonstrate that:
-
-[loweralpha]
-. The probability that a random single authentication attempt will be successful is less than one in 1,000,000; and
-. The probability that random multiple authentication attempts during a one (1) minute period will be successful is less than one in 100,000.
+The evaluator shall ensure that the TSS describes the mechanism used to generate authorization values and documents the quality metric that the mechanism provides. The information provided in the TSS shall demonstrate that the probability that a random single authentication attempt will be successful is less than one in 1,000,000.
```
jvdsn, line 1516: Following the reasoning below, I suggest: "shall demonstrate that the probability that of randomly guessing an authorization value is less than one in 1,000,000".

jvdsn (changes requested):
Just as a note for other reviewers: the concept of authentication throttling is required by FIA_AFL_EXT.1, where it fits much better.

cPP_DSC.adoc
woodbe: I don't specifically have an issue with this, I was just trying to minimize the change from what was previously specified, hence the change to only remove the throttling part of the SFR.

jvdsn: I'll open a separate PR after this one is merged

jvdsn (approved):

PR #286, opened 2024-05-02
ALC_FLR optional

### LINKED ISSUE #172: ALC_FLR addition
Once the proposal comes out for the ALC_FLR update to support EUCC/CRA, add this to the SD.

Waiting on NIAP to send this out.

woodbe:
This is to close #172 based on what the ND iTC did to add the SARs as optional. This is identical to their text except for the headings which I changes to match out style.

### DIFF
--- a/SD_DSC.adoc
+++ b/SD_DSC.adoc
@@ -3165,6 +3165,18 @@ When evaluating that the TOE has been provided and is labelled with a unique ref
 
 When evaluating the developer's coverage of the TOE in their CM system, the evaluator performs the work units as presented in the CEM.
 
+==== Basic Flaw Remediation (ALC_FLR.1) (optional)
+
+When evaluating the developer’s procedures regarding basic flaw remediation, the evaluator performs the work units as presented in the CEM.
+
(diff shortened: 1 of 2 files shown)


SD_DSC.adoc
```diff
@@ -3165,6 +3165,18 @@ When evaluating that the TOE has been provided and is labelled with a unique ref
 
 When evaluating the developer's coverage of the TOE in their CM system, the evaluator performs the work units as presented in the CEM.
 
+==== Basic Flaw Remediation (ALC_FLR.1) (optional)
+
+When evaluating the developer’s procedures regarding basic flaw remediation, the evaluator performs the work units as presented in the CEM.
```
jvdsn, line 3170: Very minor comment: you probably want to use ASCII ' instead of ’ here (and below).

jvdsn (approved):

SD_DSC.adoc
woodbe: that's what I get for a straight copy-paste :)

Updated

PR #287, opened 2024-05-02
FCS_CKM.2 reference

### LINKED ISSUE #249: Should FCS_CKM.2 be selection or optional?
The original FCS_CKM.2 was for Key establishment, but the new FCS_CKM.2 is key distribution. It doesn't seem that all modules must do this. Likely, yes, but I do not think that it is mandatory for all modules. I'm not sure how it should be placed, and how it gets pulled in exactly, since I would prefer selection-based to optional.

woodbe:
This is to close #249 by adding in the reference to FCS_COP.1/KeyEncap.

### DIFF
--- a/cPP_DSC.adoc
+++ b/cPP_DSC.adoc
@@ -691,7 +691,7 @@ _If [.underline]#Derived keys generated in accordance with FCS_CKM.5# is selecte
 
 FCS_CKM.2 Cryptographic Key Distribution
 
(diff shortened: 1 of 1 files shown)


jvdsn (approved):
I assume the SD will be updated based on the text provided by the Crypto WG

woodbe:
Yes. Once we get those updates text will be updated to match (but until then, we have what we have)

PR #289, opened 2024-05-02
RoT Test updates

### LINKED ISSUE #280: FPT_ROT_EXT.1 Tests don't make sense
The test requirements for this just state that for each of the selections to check a number of other SFRs have been successful, yet it isn't clear that some of those really satisfy this requirement.

There is text in other EAs stating that some component is checked as part of some other SFR EA, which is fine, and I can see how the RoT Storage ones make sense, but I don't quite see how the other ones link clearly (since they are more general crypto requirements or an auth one).

Is it sufficient to say these are components are tested as part of these other SFR EAs or do we really need some additional checking.

woodbe:
This is to close #280.

The changes includes several of the RoT requirements to make sure the intent is covered. I adjusted the FPT_PRO_EXT.1 to move some of the TSS stuff into the KMD as I think that is a better place for that info (I can't imagine details of the RoT being public).

I modified the language for the testing to be consistent with some other EAs which sounds better.

I added FDP_SDI.2 for FPT_ROT_EXT.2 as that is mentioned in the app note and is relevant.

### DIFF
--- a/SD_DSC.adoc
+++ b/SD_DSC.adoc
@@ -1827,9 +1827,7 @@ There are no KMD evaluation activities for this component.
 
 ====== TSS
 
-The evaluator shall ensure that the TSS describes either a pre-installed identity (contained within an SDO), or a process on how the TOE creates an identity. IEEE 802.1ar is one example of a standard which a device can use to create such an identity. 
-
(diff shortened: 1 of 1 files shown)


slpotte:
Looks mostly good to me.

slpotte:
Closing

slpotte (approved):
Found it.  Approved to merge.

PR #290, opened 2024-05-03
FPT_ITT.1 test update

### LINKED ISSUE #282: FPT_ITT.1 EA Tests
It looks like this should have an app note in the cPP that says something about expectations related to what separate parts can mean. First off there is a clear expectation that if they are part of the same board, that there is some level of protection as stated in test 3, yet tests 1 and 2 do not provide any such nuance.

Test 1 does seem to imply, since it says to follow guidance about seeing up connections, but if they are already there, it isn't clear what will be tested.

Test 2 doesn't seem to provide any explanation of when plaintext would be OK vs no interface, and if there isn't an interface, what is the test? If there isn't an interface this would seem to be a TSS or KMD activity.

While the overall intent seems clear here, the way the tests are written seem to be more problematic.

woodbe:
This is to close #282. I think with these updates that the overall testing is clear enough.

### DIFF
--- a/SD_DSC.adoc
+++ b/SD_DSC.adoc
@@ -2057,9 +2057,11 @@ Test 1: The evaluator shall ensure that all communications between distributed T
 
 Test 2: The evaluator shall ensure, for each communication channel, that the channel data is not sent in plaintext or that there is no interface by which it is possible to extract the channel data.
 
-Test 3 [conditional]: The evaluator shall, where possible, physically interrupt communications between TOE components. The evaluator shall ensure that when physical connectivity is restored, communications are appropriately protected. 
+Test 3 [conditional]: If [.underline]#modification# is selected in FPT_ITT.1.1, the evaluator shall ensure, for each communication channel, that the channel data is authenticated using an approved message authentication code, digital signature, or authenticated encryption mechanism, or that there is no interface by which it is possible to modify the channel data.
 
-Note that test 3 shall only be performed for those interfaces where physical interruption of communications is typical operational behavior. If the TOE consists of multiple components embedded within the same physical chassis such that interrupting communications requires physical destruction of the device (as opposed to disconnecting a re-connectable wire or temporarily disabling a radio), this test is not applicable.
+Test 4 [conditional]: The evaluator shall, where possible, physically interrupt communications between TOE components. The evaluator shall ensure that when physical connectivity is restored, communications are appropriately protected.
+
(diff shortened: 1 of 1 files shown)


SD_DSC.adoc
```diff
@@ -2051,11 +2053,13 @@ If applicable, the evaluator shall confirm that the guidance documentation conta
 
 The evaluator shall perform the following tests:
 
-Test 1: The evaluator shall ensure that all communications between distributed TOE components is tested during the course of the evaluation, setting up the connections as described in the guidance documentation (if applicable) and ensuring that communication is successful.
+Test 1: The evaluator shall ensure that all communications between distributed TOE components are tested. If communications channels require configuration, the guidance documentation shall be followed and ensure that communication is successful.
```
jvdsn, line 2056: "and ensure that" -> "to ensure that"?

SD_DSC.adoc
```diff
@@ -2051,11 +2053,13 @@ If applicable, the evaluator shall confirm that the guidance documentation conta
...
 
 Test 2: The evaluator shall ensure, for each communication channel, that the channel data is not sent in plaintext or that there is no interface by which it is possible to extract the channel data.
 
-Test 3: The evaluator shall, where possible, physically interrupt communications between TOE components. The evaluator shall ensure that when physical connectivity is restored, communications are appropriately protected. Note that this test shall only be performed for those interfaces where physical interruption of communications is typical operational behavior. If the TOE consists of multiple components embedded within the same physical chassis such that interrupting communications requires physical destruction of the device (as opposed to disconnecting a re-connectable wire or temporarily disabling a radio), this test is not applicable.
+Test 3 [conditional]: The evaluator shall, where possible, physically interrupt communications between TOE components. The evaluator shall ensure that when physical connectivity is restored, communications are appropriately protected.
```
jvdsn, line 2060: Is this test supposed to address the optional protection against modification that can be selected in FPT_ITT.1? If so, I don't think this test is sufficient.

There should be a test where the evaluator ensures that the TOE detects any accidental or intentional modifications to the data, if possible (i.e. if the evaluator can act as a MiTM). I believe this is different from physically interrupting the communication channel.

jvdsn (changes requested):

SD_DSC.adoc
jvdsn: @woodbe I don't think there is such a test for this specific scenario (internal data transfer). I agree with you on the test you modified, so I think that one should stay.

I added a commit that adds a new test as test 3 (for aesthetic reasons). The old test 3 is now test 4.

jvdsn: Had to force push to fix my email address...

woodbe: Ok, I think this is good

jvdsn (approved):

PR #292, opened 2024-05-06
Add alternative test to FCS_CKM.6 to address #235

jvdsn:
Address #235 by adding a new test which can be used if the volatile or non-volatile memory cannot be directly examined.

Also, Test 3 was slightly updated to change "memory" -> "the memory location" for clarity.

### DIFF
--- a/SD_DSC.adoc
+++ b/SD_DSC.adoc
@@ -484,7 +484,7 @@ The following tests require the developer to provide access to a test platform t
 
 The evaluator shall perform the following tests:
 
-*Test 1:* Applicable only if the TOE supports directly examining the SDO/SDE memory. This test is applied to each key or keying material held as plaintext in volatile memory and subject to destruction by overwrite by the TOE (whether or not the plaintext value is subsequently encrypted for storage in volatile or non-volatile memory).
+Test 1 [conditional]: If the TOE supports directly examining the SDO/SDE memory, this test is applied to each key or keying material held as plaintext in volatile memory and subject to destruction by overwrite by the TOE (whether or not the plaintext value is subsequently encrypted for storage in volatile or non-volatile memory).
 
 . Record the value of the key or keying material.
 . Cause the TOE to dump the SDO/SDE memory of the TOE into a binary file.
@@ -502,7 +502,7 @@ Steps #1.-8. ensure that the complete key does not exist anywhere in volatile me
 
 Step #9 ensures that partial key fragments do not remain in memory. If the evaluator finds a 32-or-greater-consecutive-bit fragment, then fail immediately. Otherwise, there is a chance that it is not within the context of a key (e.g., some random bits that happen to match). If this is the case the test should be repeated with a different key in Step #1. If a fragment is also found in this repeated run then the test fails unless the developer provides a reasonable explanation for the collision, then the evaluator may give a pass on this test.
 
-*Test 2:* Applicable only if the TOE supports directly examining the non-volatile memory. This test is applied to each key and keying material held in non-volatile memory and subject to destruction by overwrite by the TOE.
+Test 2 [conditional]: If the TOE supports directly examining the non-volatile memory, this test is applied to each key and keying material held in non-volatile memory and subject to destruction by overwrite by the TOE.
 
 . Record the value of the key or keying material.
 . Cause the TOE to perform normal cryptographic processing with the key from Step #1.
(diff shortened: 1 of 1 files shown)


woodbe:
It looks like this would be in place of #251, but I am not sure that the purpose of most of those changes are captured here. The earlier pull request had points about using key references to check that things were removed since it may not be possible to check memory directly. The way this one is written would seem to remove about all the possible tests if the vendor claims you can't access the memory directly.

Can you please look at https://github.com/DSC-iTC/cPP/pull/251#issuecomment-2000040169 and make sure that this is captured in this pull request? I don't see any issues with this pull request, but I think the two are somewhat in conflict as far as the intent goes. If you capture that here though, we can close the other one and apply this one instead.

Of course one thing to note here is that the Crypto WG may come back with something they want to see at some point in the SD which may override what we do here (but it may not, since we are looking at an implementation-specific SD).

jvdsn:
> It looks like this would be in place of #251, but I am not sure that the purpose of most of those changes are captured here. 

Yes, this PR would be in place of #251. To me it looks like #251 would replace the existing procedures for Test 1 and Test 2, which I don't think is the best approach.

> The earlier pull request had points about using key references to check that things were removed since it may not be possible to check memory directly. The way this one is written would seem to remove about all the possible tests if the vendor claims you can't access the memory directly.

This PR adds a test, test 4, which is performed if the memory cannot be accessed directly. I believe the steps in this PR's test 4 are equivalent to the steps from #251, but please let me know if I missed any. The reason I chose to add an extra test, instead of replacing the existing test procedures, is because I think the existing procedures are preferable if they can be performed.

> Can you please look at [#251 (comment)](https://github.com/DSC-iTC/cPP/pull/251#issuecomment-2000040169) and make sure that this is captured in this pull request? I don't see any issues with this pull request, but I think the two are somewhat in conflict as far as the intent goes. If you capture that here though, we can close the other one and apply this one instead.

I'm having a hard time interpreting the comment. Are the bullet points the result of the discussion? Because some bullet points seem to be in conflict (e.g. bullet 1 and 3). I believe this PR complies with the intent of the first bullet point, with the exception that "keys are directly accessible" -> "memory is directly accessible". This conditional is applied to all three tests though, as all three tests require the memory to be dumped or examined.

> Of course one thing to note here is that the Crypto WG may come back with something they want to see at some point in the SD which may override what we do here (but it may not, since we are looking at an implementation-specific SD).

Yes, that makes sense

woodbe:
The bullets were from a discussion and a summary. I had missed test 4 being added (my bad).

Can you make the tests [conditional] and adjust the conditions for when to use them.

Also I think we need to have some wrapping text in the SD that makes it clear that at least one of these needs to be used to be considered a pass (or something to that effect) since these would all be conditional, but as an either/or.

jvdsn:
I added the [conditional] and removed the bolding to be more consistent with the rest of the SD. I also added a note to say that each key must be tested

woodbe (approved):

woodbe:
So to be clear then, this will also close #251.

jvdsn:
@woodbe I'll check with @yiatsec if this PR addresses all her concerns. I believe it should, but I'll double check.

woodbe:
I'm sure it will be fine, I just want to make sure before we complete the merge since if we don't cover everything, then we would still need to make more edits. I think we are OK here with these changes (it covers the bullets and I think gets to the intent).

SD_DSC.adoc
```diff
@@ -517,14 +515,28 @@ Note that the primary purpose of Step #3. is to demonstrate that appropriate sea
...
 . Cause the TOE to perform normal cryptographic processing with the key from Step #1.
 . Cause the TOE to destroy the key. Record the value to be used for the overwrite of the key. 
-. Examine the memory from Step #1. to ensure the appropriate pattern (recorded in Step #3) is used. 
+. Examine the memory location from Step #1. to ensure the appropriate pattern (recorded in Step #3) is used.
+
+The test succeeds if correct pattern is found at the memory location. If the pattern is not found, then the test fails.
+
+Test 4 [conditional]: If the TOE does not support directly examining the SDO/SDE memory or non-volatile memory, this test is applied to each key and keying material held in volatile memory or non-volatile memory.
+
+. Record the value or a corresponding checksum value of the key or keying material for a given key reference.
```
yiatsec, line 529: Remove "the value or" because it is the conditional branch for TOE does not support directly examining memory and hence the key value cannot be recorded.

SD_DSC.adoc
```diff
@@ -517,14 +515,28 @@ Note that the primary purpose of Step #3. is to demonstrate that appropriate sea
...
++
+The checksum value must be deterministically computed by the TOE on the entire key or keying material. Possible methods include: error detection codes, cryptographic hashes, encryption using a fixed key.
+. Cause the TOE to perform normal cryptographic processing with the key from Step #1.
+. Cause the TOE to destroy the key.
+. Record the value (or corresponding checksum value) of the key or keying material for the key reference from Step #1.
+. Verify that the values (or corresponding checksum values) obtained from Step #1. and Step #4. are different.
+
+The test succeeds if the values (or corresponding checksum values) are found to be different. If they are found to be identical, then the test fails.
```
yiatsec, line 537: Remove "values (or" and ")".

yiatsec (approved):
For Test 4, we only need to cover verifying the checksum value because the other Test cases cover verifying the key values. Please see the inline comments at line #529 and #537.

SD_DSC.adoc
jvdsn: I was thinking about a scenario where the evaluator cannot dump the (full) memory of the TOE, but the TOE itself still has some functionality to output key values. I'm not opposed to your suggestion though.

woodbe (approved):

PR #295, opened 2024-05-08
Fix typo in FCS_CKM_EXT.3 and update reference

jvdsn:
I believe "STF" is a typo for "TSF", at least I'm not  familiar with the terminology.

Also, since the PP defines `FCS_COP.1/KeyWrap`, I think it makes sense to refer to it here.


woodbe (approved):
Nope, it's STF (Secure TOE Function) :)

PR #297, opened 2024-05-08
CC:2022 Errata update

### LINKED ISSUE #291: CC:2022 errata review
The errata needs to be reviewed for any other changes that are needed.

woodbe:
I think this is all that technically applies from the errata to the cPP (and I don't think we need anything in the SD, though I could certainly be wrong there, but with less copied directly into the SD, that is more likely).

This is to close #291

### DIFF
--- a/cPP_DSC.adoc
+++ b/cPP_DSC.adoc
@@ -4219,10 +4219,17 @@ The dependencies between SFRs implemented by the TOE are addressed as follows.
 |Rationale Statement
 
 |FCS_CKM.1 
-|[FCS_CKM_EXT.7 or FCS_COP.1] 
+|[FCS_CKM.2, FCS_CKM.5, FCS_CKM_EXT.7 or FCS_COP.1] 
+
+[FCS_RBG.1 or FCS_RNG.1]
 
 FCS_CKM.6 
-|All of FCS_CKM_EXT.7, FCS_CKM.6, and FCS_COP.1 are required by the PP.
+|FCS_CKM.2, FCS_CKM.6, FCS_CKM_EXT.7, FCS_COP.1 and FCS_RBG.1 are required by the PP.
+
+|FCS_CKM.2
+|[FDP_ITC.1, FDP_ITC.2, FCS_CKM.1 or FCS_CKM.5] 
+
+|FCS_CKM.1 is required by the PP. FDP_ITC_EXT.1 and FDP_ITC_EXT.2 are also required by the PP and equivalent to FDP_ITC.1 in this case because they all relate to a mechanism by which key data can be imported into the TSF.
 
 |FCS_CKM_EXT.7 
 |[FDP_ITC.1, FDP_ITC.2, or FCS_CKM.1] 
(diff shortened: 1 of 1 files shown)


cPP_DSC.adoc
```diff
@@ -4219,10 +4219,17 @@ The dependencies between SFRs implemented by the TOE are addressed as follows.
 |Rationale Statement
 
 |FCS_CKM.1 
-|[FCS_CKM_EXT.7 or FCS_COP.1] 
+|[FCS_CKM.2, FCS_CKM.5, FCS_CKM_EXT.7 or FCS_COP.1]
```
jvdsn, line 4222: FCS_CKM_EXT.7 is not a dependency required by CC:2022, unless I'm misreading the errata. Where does it come from?

cPP_DSC.adoc
```diff
@@ -4231,47 +4238,47 @@ FCS_CKM.6
...
 |FCS_COP.1/KAT 
-|[FDP_ITC.1, FDP_ITC.2, or FCS_CKM.1] 
+|[FDP_ITC.1, FDP_ITC.2, FCS_CKM.1, or FCS_CKM.5] 
 
 FCS_CKM.6 
 |FCS_CKM.1 and FCS_CKM.6 are required by the PP. FDP_ITC_EXT.1 and FDP_ITC_EXT.2 are also required by the PP and equivalent to FDP_ITC.1 in this case because they all relate to a mechanism by which key data can be imported into the TSF.
 
 |FCS_COP.1/KeyWrap
-|[FDP_ITC.1, FDP_ITC.2, FCS_CKM.1, or FCS_CKM.5, FCS_CKM_EXT.7, or FCS_CKM_EXT.8]
+|[FDP_ITC.1, FDP_ITC.2, FCS_CKM.1, FCS_CKM.5, FCS_CKM_EXT.7, or FCS_CKM_EXT.8]
```
jvdsn, line 4263: Again I am slightly confused why FCS_CKM_EXT.7 and FCS_CKM_EXT.8 are included here.

cPP_DSC.adoc
woodbe: I would have to check to be certain, but it was one listed in the crypto WG, so I left it.

cPP_DSC.adoc
woodbe: Same as above, I would have to go back to check, but it was already there from the crypto WG update. So that wasn't a change here, it was already included.

cPP_DSC.adoc
```diff
@@ -4231,47 +4238,47 @@ FCS_CKM.6
...
 |FCS_CKM.1 and FCS_CKM.6 are required by the PP. FDP_ITC_EXT.1 and FDP_ITC_EXT.2 are also required by the PP and equivalent to FDP_ITC.1 in this case because they all relate to a mechanism by which key data can be imported into the TSF.
 
 |FCS_COP.1/KeyedHash 
-|[FDP_ITC.1, FDP_ITC.2, or FCS_CKM.1] 
+|[FDP_ITC.1, FDP_ITC.2, FCS_CKM.1, or FCS_CKM.5] 
 
 FCS_CKM.6 
 |FCS_CKM.1 and FCS_CKM.6 are required by the PP. FDP_ITC_EXT.1 and FDP_ITC_EXT.2 are also required by the PP and equivalent to FDP_ITC.1 in this case because they all relate to a mechanism by which key data can be imported into the TSF.
 
 |FCS_COP.1/KAT
```
jvdsn, line 4256: Could you review my related PR #277? That one removes FCS_COP.1/KAT, since that is no longer defined in this cPP.

jvdsn (approved):

PR #299, opened 2024-05-08
Update for FIA_UAU.5

### LINKED ISSUE #274: FIA_UAU.5 rewrite some selections
This is low priority, but the selections don't read well if some options are selected:

FIA_UAU.5.1 - "none" does not make sense in the sentence, should probably be "no mechanism"
FIA_UAU.5.2 - the first option doesn't read well, the second one ends up with duplicated "the"

jvdsn:
Closes #274

In the SD:
* I'm not sure why the TSS EA only calls out the authentication mechanism used for the Prove service. This SFR seems more general to me, so the TSS should include all authentication mechanisms. Also the wording "how each authentication mechanism provides authentication" doesn't really make sense to me, so I changed it to "how each mechanism enforces the authentication".
* AGD EA: same as the TSS, why would only the Prove service be listed here?
* Test EA: "no interface to provide authentication credentials over an external interface" says the same thing twice (interface), so I just removed "over an external interface"

In the cPP:
* For FIA_UAU.5.1, change "none" to "no mechanism".
* For FIA_UAU.5.2, try to work around the awkward sentence construction from CC. I'm not sure if this will render correctly though.

### DIFF
--- a/SD_DSC.adoc
+++ b/SD_DSC.adoc
@@ -1594,15 +1594,15 @@ There are no KMD evaluation activities for this component.
 
 ====== TSS
 
-The evaluator shall examine the TSS and ensure that it describes the authentication mechanisms used to support user authentication for the Prove service as well as how each authentication mechanism provides authentication for the Prove service.
+The evaluator shall examine the TSS and ensure that it describes the authentication mechanisms used to support user authentication, including how each mechanism enforces the authentication.
 
 ====== AGD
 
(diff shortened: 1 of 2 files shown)


woodbe (approved):

woodbe:
I think this works. I made a minor adjustment, but nothing that impacts the change. I agree the working is awkward.

Issue #35, opened 2022-12-20
CCDB AU/005 - Section 3 CC Conformance Claims


woodbe:
### Comment:
The text claims conformance with APE_OBJ.2 and APE_REQ.2.  This implies a lot of work with mapping objectives to the SPD and mapping the SFRs to the TOE objectives.

### Proposal:
Consider the NDcPP approach where APE_OBJ.1 and APE_REQ.1 are used – requiring far less wordy mapping and rationale work.

### Priority: 2

woodbe:
This would seem to be a significant change. I am not advocating either way, but this is not something that can be handled by a TD, which by definition would mean that it needs to be a v1.1 change.

woodbe:
need to check if this will be applicable in #125.

jdonndelinger:
So CC2022 allows for PPs to be written without the objectives section. From Part 1. page 158 0f 167:
"D.4 Direct Rationale STs
D.4.1 General
In some situations, it is appropriate to omit the definition of the TOE security objectives. In this case the Security Requirements rationale directly maps the SFRs and, where appropriate, security objectives for the operational environment, to the SPD.
The intention of the Direct Rationale ST is to minimize the level of indirection between the SPD, any security objectives for the operational environment, and the SFRs, based on an enhanced description of the SFRs.
The differences found in a Direct Rationale ST are in the conformance claims, security objectives and in the SPD sections. These are described in D.4.2 and D.4.3, below."

When we wrote the ND and FDE cPPs we used this approach, it was called something different (Low Assurance) in the earlier versions of the CC. 

We can adopt this approach as long as we are in keeping with CC2022. We'll just have to make some editorial changes. This should not affect the SFRs (cPP) or EAs (SD) in any way. It will streamline the cPP and reduce work on the iTC and evaluations (less requirements and work units in evaluating ST).

woodbe:
As you can't mix the two types together, need to see what NIAP is planning for their updates to CC:2022.

woodbe:
NIAP has stated that this needs to be done for this release

PR #310, opened 2024-06-19
Direct rationale PP

smuellerDD:
As specified in https://github.com/DSC-iTC/cPP/issues/35 the approach of the direct rationale PP/ST shall be taken. The change removes the objectives for the TOE and replaces the objectives to SFR mapping with the SPD to SFR mapping as defined in CC2022 Appendix Part 1 B.5.

The approach to the mapping was to apply the threads mapped to objectives when replacing the objectives in the SFR mapping table.


jvdsn:
This will likely need a rebase or merge if #308 gets merged first because of the changes in the SFR rationale.

jvdsn:
I was reading through the newly endorsed AppSW cPP, which is a "low assurance PP", and they do it a little differently.

woodbe (approved):

PR #319, opened 2024-07-10
Add allowance to use PBKDF for password hashing

### LINKED ISSUE #313: PBKDF2 question to be resolved
https://github.com/DSC-iTC/cPP/pull/308#discussion_r1629833539

Review and discuss whether the DSC should allow for more options than are allowed by the crypto catalog.

jvdsn:
Closes #313 

A couple of references to PBKDF2 being used for password hashing purposes:

* OWASP: https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html
* NIST SP 800-63B: https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-63b.pdf ("Key derivation functions take a password, a salt, and a cost factor as inputs then generate a password hash. Their purpose is to make each password guessing trial by an attacker who has obtained a password hash file expensive and therefore the cost of a guessing attack high or prohibitive. Examples of suitable key derivation functions include Password-based Key Derivation Function 2 (PBKDF2) [SP 800-132] and Balloon [BALLOON].")

I'm open to feedback.

### DIFF
--- a/cPP_DSC.adoc
+++ b/cPP_DSC.adoc
@@ -2651,9 +2651,11 @@ _If deriving a secret IV or seed, select KDF-HASH, KDF-MAC-1S, or KDF-MAC-2S._
 
 FCS_CKM_EXT.8 Password-Based Key Derivation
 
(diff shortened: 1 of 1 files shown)


slpotte (approved):
For this PP, this is reasonable.  For the catalog, I would prefer we put this in a parking lot and visit this during the next revision.  The primary reason is that we really want to finish the catalog and publish it.

woodbe (approved):

PR #321, opened 2024-07-18
PP structure is in Annex B, not Annex C

jvdsn:
Original comment:

Don't understand why there is a reference to [CC1, Section C.1].

CC:2022 Annex C talks about the structure of  PP-Modules and PP-Configurations. Instead, CC:2022 Section 6.3.3.2 summarizes the information delivered in a  Protection Profile.  Moreover, CC:2022 Annex B talks about the structure of Protection Profiles.

### DIFF
--- a/cPP_DSC.adoc
+++ b/cPP_DSC.adoc
@@ -52,7 +52,7 @@ The DSC international Technical Community (iTC) designed the DSC cPP as a standa
 
 === Scope of Document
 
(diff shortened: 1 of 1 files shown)


woodbe (approved):

woodbe:
I would assume that in 3.1R5 it makes sense, but changing to 2022 it doesn't. I definitely hadn't looked for those types of nuances when updating.

PR #329, opened 2024-07-18
Selection should be underlined rather than italic

jvdsn:
Original comment:

Editorial.

A selection operation has been performed on this SFR to get the following text

[deallocation of the resource from] in the SFR.

The text style should be underlined rather than italicized, because this is a selection.


woodbe:
How does this relate to #334?

woodbe (approved):

PR #336, opened 2024-07-18
FCS_CKM.5 app note edit

woodbe:
The first paragraph of the app note in the catalog is really intended for people using the catalog, not for inclusion in the cPP.

### DIFF
--- a/cPP_DSC.adoc
+++ b/cPP_DSC.adoc
@@ -2647,9 +2647,7 @@ NIST SP 800-108 Rev. 1 (Section 4.3) [KDF in Double-Pipeline Iteration Mode] [se
 
 |===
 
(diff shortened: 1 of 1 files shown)


jvdsn (approved):

slpotte:
@woodbe, there has been a notion that we would provide commentary/guidance on how to use the catalog.  The is a long-time goal especially of @jdonndelinger.  I think some of the commentary/guidance wound up as app notes.  I would defer to @jdonndelinger on how to best separate legitimate app notes from commentary/guidance.

woodbe:
I agree, we need to do a check on things that should be removed, I noticed this particular one as obvious.

Maybe in the catalog you move those things out of app notes and into something above the requirements (or below but noted as "PP/ST author notes"

slpotte:
Yes, some were moved to introductions.  There are also sections labeled "Guidance" under FCS_CKM.2 and FCS_COP.1/Hash.  It seems inconsistent to provide some guidance in the introductions, and then more under these labels.  Feel free to remove "guidance" from the app notes.  We will clean up the catalog later.

PR #339, opened 2024-07-18
Update cPP_DSC.adoc

### LINKED ISSUE #309: [cPP PUBLIC REVIEW]
Section 2.2, End of third paragraph.

**Comment Type.**
S = Substantial: a comment or issue significant enough to warrant a possible objection if the change (or a similar objective) is not met.

The statement about SoC makes it sound like the TOEs must be SoCs, not just that they could be. 

 Suggest deleting “are integrated into a System on Chip (SoC) and as such”.

woodbe:
This is to close #309

### DIFF
--- a/cPP_DSC.adoc
+++ b/cPP_DSC.adoc
@@ -171,7 +171,14 @@ image::representationofTOE.png[]
 
 The TOE should be one or more discrete and embedded hardware components that provide well-scoped security functions that are physically inaccessible directly from the rich operating system. The DSC TOE would consist of isolated firmware and circuitry capable of executing well-defined commands against SDEs/SDOs within the TOE and outside the TOE across restricted interfaces. The DSC TOE is not intended to be a discrete, separate stand-alone component, but one which is directly embedded into a larger system.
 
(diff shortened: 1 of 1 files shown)


cPP_DSC.adoc
```diff
@@ -171,7 +171,14 @@ image::representationofTOE.png[]
...
 
-A DSC may be comprised of a single embedded component within a device, such as a Secure Enclave Processor (SEP), while in other cases it may be a multi-component system comprised of a software layer and several hardware components (which may be discrete or embedded), such as a Trusted Execution Environment (TEE). Other configurations are possible, with the key point being the DSC is embedded within a larger system and is not a discrete component. These dedicated hardware/software components are integrated into a System on Chip (SoC) and as such are isolated components of a larger physical package. <<ExofTOEInt>> below shows a block diagram of a typical example of a DSC TOE with all of its internal components.
+There are many possible configurations for a DSC. Some examples are:
+
+* A DSC may be comprised of a single embedded component within a device, such as a Secure Enclave Processor (SEP) or System on Chip (SoC)
```
jvdsn, line 176: I'm not sure if the DSC itself would actually be the SoC, as opposed to be integrated in the SoC. Below (and above) we say that the DSC is not a separate stand-alone component, whereas SoCs are generally stand-alone?

woodbe: I think the point here was that it could be. This is just an example, and while an eSE shouldn't be evaluated to the DSC cPP, it could, and so should not be excluded.

slpotte (approved):

PR #340, opened 2024-07-18
FCS_CKM_EXT.7 app note

### LINKED ISSUE #239: Add phrase to FCS_CKM_EXT.7.1 App Note 5

woodbe:
This is to close #239

### DIFF
--- a/cPP_DSC.adoc
+++ b/cPP_DSC.adoc
@@ -774,7 +774,7 @@ _FCS_CKM.5 defines KDF-CTR, KDF-FB, and KDF-DPI._
 +
 _For the KDF functions, when concatenating keys for AES-CMAC, the contributions from each party should be an equal number of bits, resulting in the selected key size (e.g., if each share is 128 bits and there are two parties, then the result after concatenation is a 256-bit key, which is appropriate only for AES-256-CMAC). For HMAC functions, the shares can be any size as long as the concatenated result is equal to or greater than the nominal cryptographic strength of the chosen hash function (e.g. if each share is 128 bits, then the result after concatenation is 256 bits, which can be used in any of SHA-1, SHA-256, or SHA-512)._
 +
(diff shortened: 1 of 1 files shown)


cPP_DSC.adoc
```diff
@@ -772,9 +772,9 @@ _Application Note {counter:remark_count}_:: _This SFR captures methods for multi
...
 +
-_For the KDF functions and XOR, each party may have to use an asymmetric method from FCS_CKM_EXT.7 to transmit their shares to each other. Key shares may also come from a token, in which case, TOEs may use key access methods in FCS_CKM.3 to authorize access and use of those keys in this SFR._
+_For the KDF functions and XOR, each party may have to use an asymmetric method from FCS_CKM.2 to transmit their shares to each other.  Key shares may also come from a token, in which case, TOEs may use key access methods in FCS_CKM.3 to authorize access and use of those keys in this SFR._
```
jvdsn, line 777: GitHub tells me you added an extra space between "other.  Key shares"?

jvdsn (changes requested):

cPP_DSC.adoc
woodbe: I'm sure it happened because I literally just copied this from the doc (where there were double spaces) and pasted it in without checking. I'll do a quick check for any while clearing this one.

woodbe: of course a full search/replace only found that one

jvdsn (approved):

slpotte (commented):
Looks fine to me.

PR #345, opened 2024-07-24
Make ECD formatting consistent with CC:2022

### LINKED ISSUE #334: [cPP PUBLIC REVIEW] atsec21 - extended component definition formatting
Appendix C

**Comment Type.**
A = Administrative: the issue is not significant enough to warrant an objection to the cPP if it isn't changed.

Editorial. As to formatting conventions, it is stated in cPP that "formatting conventions for selections and assignments in this Appendix are those in [CC2]."

In both CC v3.1 and CC:2022, the assignment and selection operations are presented in italicized text. Some extended SFRs defined in cPP Appendix C, e.g., the FCS_STG_EXT.1, FDP_ITC_EXT.2, use underlined texts.

Investigate if the formatting needs to match CC2022 and if so, change the ECD.

jvdsn:
The rules for formatting in CC:2022 seem to be as follows:
- Italic is used for both assignment and selection
- The prefix "assignment: " and "selection: " are not formatted at all, nor are the surrounding brackets []
- For nested assignments / selections, the whole nested entry is in italic

Also fixes a forgotten selection option for FCS_CKM_EXT.8 (PBKDF) output length.

Closes #334

### DIFF
--- a/cPP_DSC.adoc
+++ b/cPP_DSC.adoc
@@ -3146,7 +3146,7 @@ FCS_COP.1/KeyedHash Cryptographic Operation (Keyed Hash) +
 FCS_CKM.6 Timing and event of cryptographic key destruction 
 
 [vertical]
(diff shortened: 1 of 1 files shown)


jvdsn:
I have no idea why some entries are prepended with "{empty}".

woodbe:
The {empty} is one of those quirks of asciidoc. Basically if you take it out, when the [] get processed with the underlines and such, things break. I had been using some other odd constructs to make it work then found that one, and it worked cleaner, so that is what they are for.

woodbe:
I removed the {empty} brackets as they were needed specifically for cases where you have underlines inside the [] as the markers for all that stuff cause problems. Since those are removed, then they could be safely removed.

woodbe (approved):

slpotte (approved):

PR #348, opened 2024-08-07
Use latest revision of IEEE standard for XTS

jvdsn:
Two other instances of XTS (FCS_COP.1.1/SKC) refer to the 2018 revision, so I don't see any reason to refer to the old revision here.


woodbe (approved):

slpotte (commented):
I made the change to the catalog today.

PR #350, opened 2024-08-07
App notes

### LINKED ISSUE #341: Crypto catalog app note review
Some of the app notes from the catalog are more for the ":authors" and not for the document itself. We need to review the app notes in the cPP to verify if the ones in there for the FCS requirements are appropriate or not.

jvdsn:
I did not go through all app notes, but just took out the obvious issues by searching for some keywords. Perhaps a more thorough search is required before merging this.

Closes #341.

### DIFF
--- a/cPP_DSC.adoc
+++ b/cPP_DSC.adoc
@@ -1226,7 +1226,7 @@ in accordance with [assignment: _list of standards_].
 
 _Application Note {counter:remark_count}_:: _No rationale is acceptable for not satisfying one of these dependencies._
 +
-_If a reseeding is selected in the first selection and something other than “never” is selected in the third selection, but reseeding is not feasible, the TSF will uninstantiate RBGs, rather than produce output that is of insufficient quality. The listed standards should specify the reseed interval and procedure for uninstantiating and reseeding. The remaining selection allows the PP Author to require application-specific conditions for reseeding._
+_If a reseeding is selected in the first selection and something other than “never” is selected in the third selection, but reseeding is not feasible, the TSF will uninstantiate RBGs, rather than produce output that is of insufficient quality. The listed standards should specify the reseed interval and procedure for uninstantiating and reseeding._
 +
 _"Uninstantiate" means that the internal state of the DRBG is no longer available for use._
 +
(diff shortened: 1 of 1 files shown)


woodbe:
In app note 11, do we need to have the descriptions as to what salts, nonces and IVs do? This seems like something that could be removed given this is a PP on DSC, not a cryptographic teaching document.

woodbe:
I made some minor edits on my read through. I think this is ready to be reviewed for merging.

jvdsn:
I can't approve my own PR, but your changes look good to me.

woodbe (approved):

slpotte:
I'm not sure of the status of this is in DSC, but Bob Clemons and I are thoroughly scrubbing the app notes in the whole catalog.

woodbe:
We think we are done, but if something new should be removed, we can do so of course.

slpotte:
@JDavid-Thompson This component was conceived by the full drive encryption working group.  The goal was to set requirements for full drive encryption on wear-leveling technologies.  It's a really good idea to turn on the encryption feature of self-encrypting drives, especially those that use wear-leveling technologies.  It's a bad idea to store private keys on this type of technology without the encryption turned on.  With encryption on, by erasing the media encryption key (MEK), which presumably is NOT stored in wear-leveling technology (or is encrypted if it is), then any private keys stored on the encrypted drive are effectively "cryptographically erased" according to NIST standards.  In practice, the MEK should be stored somewhere other than these wear-leveling technologies; I'd have to go back and check, but this seems like a reasonable assumption.  In practice, MEKs are stored as encrypted blobs, probably in a chain of encrypted blobs, leading to a root that is not encrypted, but certainly not stored in a wear-leveling technology.  

For wear-leveling technologies, the assumption is that the encryption feature is turned on, and by deleting the MEK (or some other intermediate key downstream from the MEK) that the private keys stored there are cryptographically erased, and that action satisfies this requirement.

If one insists on storing private keys in wear-leveling technologies with the encryption feature turned off, I don't see how we can write a requirement to ensure with confidence that all copies of a private key are erased across all the proprietary implementations of this technology.  There has to be onus put on the designers of products to correctly plan the storage of private keys so that they are not unwittingly exposed on wear-leveling technologies.

woodbe:
Please continue any discussion on the wear leveling in #229 as that is the place where this is being worked on, not here in a closed pull request on app notes.

PR #353, opened 2024-08-12
Proposed Crypto EA update

woodbe:
This is a proposed change for the Evaluation Activities for crypto algorithms.

I have kept this to a single SFR, FCS_COP.1/Hash to minimize the amount of changes to look at. I think this is written sufficiently clear as to know what to do, while being sufficiently vague as to be able to used repeatedly with minimal changes for all the crypto requirements.

Once the proposal is sufficient so as to be agreed to, I will copy it to the rest of the requirements below. This will probably be as individual pull requests (so the minor modifications are easier to see individually).

The list of SFRs that I see needing this is (in the order they are in the doc):

FCS_CKM_EXT.7
FCS_COP.1/Hash
FCS_COP.1/KeyedHash
FCS_COP.1/SigGen
FCS_COP.1/SigVer
FCS_COP.1/SKC
FCS_COP.1/AEAD
FCS_COP.1/CMAC
FCS_COP.1/KeyEncap
FCS_COP.1/KeyWarp
FCS_CKM.5
FCS_CKM_EXT.3 (new) - This didn't need the separate algorithm testing so it is in #361 
FCS_CKM_EXT.8 (new)


SD_DSC.adoc
```diff
@@ -597,67 +597,22 @@ The evaluator checks the AGD documents to determine that any configuration that
...
+* input length
+* optional input variables 
+* output length
+* bit vs byte orientation
 
-*Pseudo-randomly Generated Messages (Monte Carlo) Test, Byte-oriented Mode*
+The evaluator shall generate a set of test vectors to exercise the algorithm according to the specified implementation details. Any parameters with a range shall test, at least, the minimum and maximum values specified for the parameters. The test vectors shall test each possible combination of the parameters. For example, if the algorithm has 2 optional input variables, a range of input lengths and 1 output length, then for each input length that is tested, each configuration of the optional input variables shall be tested (if the variables can be used together, that is another combination to be tested).
```
jvdsn, line 613: This might be possible for some simple algorithms (e.g. AES with each padding mode, each key size, and maximum payload lengths). However, as more parameters become available, this becomes infeasible.

For example, for SP 800-56Br2 key transport (aka RSA OAEP), at least the following parameters exist:
- Role (initiator / responder)
- Key generation method (rsakpg1-basic, rsakpg1-prime-factor, rsakpg1-crt, rsakpg2-basic, rsakpg2-prime-factor, rsakpg2-crt)
- Modulo (2048, 3072, 4096, 6144, 8192 bits)
- Length of the key to derive (128-1024 bits)
- Hash algorithm (any of SHA-1, SHA-2, or SHA-3)
- Associated data pattern (if applicable)
- Key confirmation (a whole set of additional information)

SP 800-56Ar3 and SP 800-56Br2 define algorithms with even more parameters / combinations. The NIST ACVP server will not generate test vectors for all possible combinations. Instead, the server will generate 20 "test groups" that each test a random combination, and for each test group, 10 test vectors are included.

I suggest changing the sentence to "The test vectors shall test a reasonable number of combinations of the parameters to ensure coverage of the entire algorithm implementation."

jvdsn (changes requested):

SD_DSC.adoc
jvdsn: The following sentence should probably also be changed then, to "For example, if the algorithm has 2 optional input variables, a range of input lengths and 1 output length, then for each input length that is tested, each configuration of the optional input variables should be tested (if the variables can be used together, that is another combination to be tested)."

woodbe: @jvdsn Thank you. This is what I needed (since while I know how to do CAVP testing, I don't actually know what is going on underneath at the detailed level).

I have posted a new commit to try and bring this out.

It is possible we will need to tweak things based on the type of algorithm, which is annoying, but fine.

jvdsn:
I'll go through the test vectors from the server for a few more algorithms to confirm what I'm seeing. I'll provide more feedback later today.

woodbe:
The "complexity" of the algorithm should also include parameters in the TOE algorithm to determine the number of test groups that is reasonable.

Should a lower number be specified?

> Test all combinations of parameters, or if infeasible, test 20 random combinations (with 10 test vectors each combination) to perform a reasonable review of the algorithm

woodbe:
Create test types for non KAT (i.e. Monte Carlo or similar). Then plug and play these tests into the algorithm types where we need them.

jvdsn:
Some notes while I'm reviewing the ACVP test vectors:
- For block ciphers, the number of KAT test groups varies between 24 and 30, whereas the number of test cases in a test group varies between 7 and 128.
- For block ciphers, MCT test cases are only provided for CBC, CFB1, CFB8, CFB128, ECB, and OFB modes. I am not sure why. There is one MCT test case for each key size and input key length (i.e., 6 MCTs if encrypt/decrypt for [128, 192, 256] bits).
- For SHA-1 / SHA-2, there is one KAT test group containing 512 test cases. There is also one MCT test group containing one test case.
- For SHA-3, there's one KAT test group containing 174 test cases. There is also one MCT test group containing one test case.
- For HMAC, there are 13 KAT test groups containing 75 test cases each. Each test group has a specific (message) input size, key size, and MAC tag size.
- For KMAC, there is one KAT test group for each combination of xof and hex customization, containing 100 test cases each. Here, the input key size is defined at the test case level, not the test group level.
- For KMAC, there is one MVT (MAC Verify Test) test group for each combination of xof and hex customization, containing 100 test cases each. Here, the input key size is defined at the test case level, not the test group level.  
- For PBKDF, there's one test group for each supported algorithm, containing 10 test cases each. Here, the output key size is defined at the test case level, not the test group level.
- For CTR_DRBG, each algorithm has 10 test groups: 4 with prediction resistance = false and derivation function = true, one with prediction resistance = false and derivation function = false, 4 with prediction resistance = true and derivation function = true, and one with prediction resistance = true and derivation function = false. Each test group has 15 test cases.
- For Hash_DRBG and HMAC_DRBG, each algorithm has 8 test groups: 4 with prediction resistance = false and 4 with prediction resistance = true. Each test group has 15 test cases.
- For ECDSA SigGen/SigVer, there's one test group for each combination of curve and hash. There are 21 test cases in each test group.
- For ECDSA KeyGen, there's one test group per curve. There are 10  test cases in each test group.
- For RSA SigGen, there's one test group for each combination of modulus size, hash, and padding type. There are 10 test cases in each test group.
- For RSA SigVer, there's 3 test groups for each combination of modulus size, hash, and padding type. There are 6 test cases in each test group.
- For RSA KeyGen, there's one test group per modulus size. There are 25 test cases in each test group.
- For FFCDH, there's one test group per DH group. Depending on the scheme and KAS role (randomly chosen for the test group), there's 10-15 test cases.
- For ECDH, there's one test group per curve. Again, depending on the scheme and KAS role (randomly chosen for the test group), there's 10-15 test cases.

Of course, there's many more algorithms, I just looked at a selection. All that being said, there's not much consistency in ACVP. There seems to be at least 6 test cases in a test group for the "normal" (KAT) tests, but only one test case for the MCTs.

woodbe:
I'm good with the testing laid out this way. It is more prescriptive for the types of test, specifies nuances per algorithm where needed, and seems like a good balance.

jvdsn:
@woodbe if the general approach is accepted during the meeting, I'll add more of them for the other SFRs.

jvdsn:
One other thing I found out regarding ACVP is that the MCT isn't always consistent. In fact, the MCT that ACVP has for SHA-1/SHA-2 is different from the one for SHA-3. Also, the block cipher MCT descriptions in existing PPs don't correspond exactly to the one implemented by ACVP. I don't know if NIAP overlooked this, or if they decided that ACVP was "close enough".

woodbe:
Add test cases and test groups to the glossary at a high level. Test cases will be defined for each SFR on its own.

woodbe:
It is agreed at this point to proceed with this type of testing for publication, and then see when the Crypto WG publishes their EAs. If schemes submit comments on the testing, they will be handled at that time with a future update (as would integration of the Crypto WG EAs).

jvdsn:
I have done:
FCS_CKM_EXT.7
FCS_COP.1/Hash
FCS_COP.1/KeyedHash
FCS_COP.1/SigGen
FCS_COP.1/SigVer
FCS_COP.1/SKC
FCS_RBG.1
FCS_CKM.1/AKG
FCS_CKM.1/SKG

Still open:
FCS_CKM.5
FCS_CKM_EXT.8
FCS_COP.1/AEAD
FCS_COP.1/CMAC
FCS_COP.1/KeyEncap
FCS_COP.1/KeyWarp

jvdsn:
The remaining SFRs should all be addressed now. Please double-check to make sure I did not forget anything. A couple of notes:
- For PBKDF, there was a bunch of outdated information in the EAs, including references to a requirement for at least 1000 iterations. I believe that requirement was removed, so I also removed that part of the EA.
- For FCS_CKM.5, there are specific requirements on TwoStep KDFs:
** If an AES-CMAC (with key length 128, 192, or 256 bits) is used in the randomness extraction step, then AES-CMAC with a 128-bit key is used as the PRF in the key expansion step.
** If an AES-CMAC is being used as the MAC, the salt length shall be the same length as the AES key (i.e. 128, 192, or 256 bits).
These requirements are directly from the NIST standard. However, in the DSC we also allow CMAC to be used in the KDF with other block ciphers (e.g. HIGHT, LEA). The NIST standard doesn't define any rules for those ciphers. Do we want to expand NIST's requirements? Or keep it only for AES: then if someone does select one of the other block ciphers, they are not bound by these requirements?

woodbe:
> RE:  If an AES-CMAC is being used as the MAC, the salt length shall be the same length as the AES key (i.e. 128, 192, or 256 bits).

> These requirements are directly from the NIST standard. However, in the DSC we also allow CMAC to be used in the KDF with other block ciphers (e.g. HIGHT, LEA). The NIST standard doesn't define any rules for those ciphers. Do we want to expand NIST's requirements? Or keep it only for AES: then if someone does select one of the other block ciphers, they are not bound by these requirements?

I think for simplicity we just keep the general rules and expect that those other ciphers will work the same. If they don't, then we can handle it via a TD at this point.

woodbe:
I will leave this open until tomorrow in case anyone wants to make a last review/edit. Then I will force the merge of this Pull Request and publish the updated release tomorrow.

jvdsn:
> I think for simplicity we just keep the general rules and expect that those other ciphers will work the same. If they don't, then we can handle it via a TD at this point.

Right, the only reason I didn't do that just yet is because I'm not very familiar with those ciphers. I think some of them might have different block / key sizes than AES so the rules might be analogous but not identical.

woodbe:
I think we are OK leaving it as it is for now though unless you think there is something more you can do that would provide flexibility in that (I suspect that for the level of detail we are doing we are fine though).

woodbe:
With no additional comments, I am going to merge this per agreement for publication as the Proposed Draft.

PR #363, opened 2024-09-06
Publication date/version

woodbe:
This is for the final publication date and version.

I am removing the interim releases for the drafts as those aren't needed for the final publication but are useful for the draft release process.

### DIFF
--- a/SD_DSC.adoc
+++ b/SD_DSC.adoc
@@ -5,8 +5,8 @@
 :sectnums:
 :sectnumlevels: 5
 :imagesdir: images
-:revnumber: 2.0-Proposed-Draft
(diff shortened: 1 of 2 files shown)


jvdsn:
What about the Proposed Update stage? On the image at https://dsc-itc.github.io/

woodbe:
That stage is when we make the updates on comments that come in from this period. So basically the assumption is that we don't get many comments on the proposed draft round, so it doesn't take long to resolve them and publish.

Basically that is 10 days after the comment period closes to make those changes, and that is the final version. 

So this PR is based on that assumption for timing, at the end of the period

jvdsn (approved):

PR #373, opened 2024-10-07
FCS_CKM_EXT diagram

### LINKED ISSUE #366: [cPP PUBLIC REVIEW] [atsec04] Family diagram in C.1.1. is inaccurate
The diagram in C.1.1. says "FCS_STG_EXT", however the actual family is "FCS_CKM_EXT"

woodbe:
This is to close #366

### DIFF
--- a/cPP_DSC.adoc
+++ b/cPP_DSC.adoc
@@ -3080,7 +3080,7 @@ This family defines requirements for key life cycle operations.
                                                     |  +---+
     +--------------------------------------------+  |
     |                                            |  |  +---+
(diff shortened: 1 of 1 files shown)


jvdsn (approved):
I didn't realize the diagram itself was also generated by adoc, that makes things a lot easier.

woodbe:
Yup. It's called ditaa. It's very simple (there are more complicated systems that may be better, but this was the one I was able to get to work consistently). The thing I like is that the diagram is basically old ascii art, but it means it can be read in the raw file as well as generated image output that looks basically like what is in the CC docs.

PR #374, opened 2024-10-07
FCS_CKM_EXT.3 hidden requirement

### LINKED ISSUE #370: [cPP PUBLIC REVIEW] [atsec05] FCS_STG_EXT.1 and FCS_CKM_EXT.3
Original comment:
```
1. Application Note 12 on Page 50 states:

If software-based is selected, the ST author is expected to select all software-based key storage in FCS_CKM_EXT.3.

However, FCS_CKM_EXT.3 does not provide such an selection option.

2. In addition, FCS_CKM_EXT.3 is a selection-based SFR. I cannot find info explicitly indicating when to claim it in the ST.
```

Maybe this application note should be "If software-based is selected, the ST author must include FCS_CKM_EXT.3."

woodbe:
This is to close #370.

The logic here is that selecting the software storage does not explicitly mandate the use of FCS_CKM_EXT.3, but that is the intent. Having this as an app note is a hidden requirement and is not allowed. By adding in the statement that protection of software-based storage has to be according to that SFR, it makes that requirement visible and explicit.

With that I don't think we actually need to have the app note anymore, since it really doesn't have any reason to be there. We do have this in some other places, but when it is explicit, I think we can drop it.

### DIFF
--- a/cPP_DSC.adoc
+++ b/cPP_DSC.adoc
@@ -1317,9 +1317,7 @@ _Each algorithm and mode have varying guidance on the lengths of the salts, nonc
 
 FCS_STG_EXT.1 Protected Storage
 
-FCS_STG_EXT.1.1:: The TSF shall provide {empty}[selection: [.underline]#mutable hardware-based, immutable hardware-based, software-based]# protected storage for {empty}[selection: [.underline]#asymmetric private keys, symmetric keys]# and {empty}[selection: [.underline]#persistent secrets, no other keys]#.
-
-_Application Note {counter:remark_count}_:: _If [.underline]#software-based# is selected, the ST author is expected to select [.underline]#all software-based key storage# in FCS_CKM_EXT.3._
+FCS_STG_EXT.1.1:: The TSF shall provide {empty}[selection: [.underline]#mutable hardware-based, immutable hardware-based, software-based in accordance to FCS_CKM_EXT.3]# protected storage for {empty}[selection: [.underline]#asymmetric private keys, symmetric keys]# and {empty}[selection: [.underline]#persistent secrets, no other keys]#.
 
 FCS_STG_EXT.1.2:: The TSF shall support the capability of {empty}[selection: [.underline]#importing keys/secrets into the TOE, causing the TOE to generate keys/secrets]# upon request of {empty}[selection: [.underline]#a client application, an administrator]#.
 


slpotte:
I can't see the list of dependencies in this snippet.  This intent is a reminder to include the relevant components for software-based key wrapping.  FCS_CKM_EXT.3 may not have been the best choice there.  FCS_COP.1/KeyWrap may be a better choice.

woodbe:
The reason I pointed to the FCS_CKM_EXT.3 is that if you choose to do software key storage, that one provides the options that you may use to do the protection claims. So instead of telling the user in multiple locations to link to FCS_COP.1/KeyWrap (or KeyEncap or AEAD), I point them to that SFR, which then points to the FCS_COP.1. To me anyway, this establishes the chain of the requirements and how they link together. I do get having them point directly, but the point of FCS_STG_EXT.1.1 is to say how you are doing the protection, and the option of SW forces you to bring in FCS_CKM_EXT.3 to then specify how you do the SW protection. If the SFR points instead to the other requirements directly, then FCS_CKM_EXT.3 is probably unnecessary (or maybe not, but I don't see how it actually gets readily linked in when it is actually needed).

jvdsn:
> The reason I pointed to the FCS_CKM_EXT.3 is that if you choose to do software key storage, that one provides the options that you may use to do the protection claims. So instead of telling the user in multiple locations to link to FCS_COP.1/KeyWrap (or KeyEncap or AEAD), I point them to that SFR, which then points to the FCS_COP.1. To me anyway, this establishes the chain of the requirements and how they link together. I do get having them point directly, but the point of FCS_STG_EXT.1.1 is to say how you are doing the protection, and the option of SW forces you to bring in FCS_CKM_EXT.3 to then specify how you do the SW protection. If the SFR points instead to the other requirements directly, then FCS_CKM_EXT.3 is probably unnecessary (or maybe not, but I don't see how it actually gets readily linked in when it is actually needed).

I am also wondering if we're not having one too many links in this chain. Maybe FCS_CKM_EXT.3 is superfluous. In any case, I think this PR does fix the immediate issue.

jvdsn (approved):

woodbe:
Decided that https://github.com/DSC-iTC/cPP/pull/374#issuecomment-2405540002 will not be done.

Issue #381, opened 2024-10-10
[cPP PUBLIC REVIEW] [atsec13] Application Note consistence and usage of "should"


jvdsn:
> Application Note 27 lists several details the ST "should" include, for example, the policies allowed and how the TOE evaluates them; location of these policies and how the TOE protect their integrity, etc. However, the EAs in section 2.3.3.3.1 only requires the TSS to describe each of the options for reauthorization.

This issue is twofold:

1) There are a lot of instances in the cPP where we say "the ST [author] should" do something. According to the CC, "should" is optional if a rationale is provided. Are we sure that every one of those instances is actually optional? For example: "TSFs that employ other algorithms or modes that require OTVs should include FCS_OTV_EXT.1."

2) If the cPP says "the ST [author] should" or "the ST [author] must" do something, does it need to be added to the TSS evaluation activities for that SFR? I would think so, in general.

woodbe:
Take out the last 2 sentences in the paragraph for point 1.

woodbe:
The scope of the change will determine whether this is accepted now or later. This review cycle is for editorial changes, so there is a limit to what we can accept at this point. 

Consistency of the language for point 2 (especially on the "must" text) would definitely be editorial. Changes to the SD based on that are more questionable, but if it is adding a "check this" for maybe 5-6 SFRs, it should be fine, but more than that and the number of changes is a little too much.

@jvdsn please create a new Issue for the "should" issue, tag this one as reference and give it a v3 target.

laurie-tyz:
Just a thought here... In the International Standards Organization, we follow this wording:
     ISO term -------> meaning
Shall, shall not - must, required
Should, should not - recommended
 may, need not - permission
can, cannot - possibility or capability

woodbe:
Totally agree, it's one of our cleanup issues, and that is what I am expecting to use. I also have been doing work in ETSI and they won't even allow "must" as a term, which we have a lot in here. Not that this is something we need to change, it's more that we need to actually be consistent (there have been a lot of different hands in this over the years and hence styles).

laurie-tyz:
Yes...  thanks.  The wording I suggested stands to this day.  One of the first things those of us commenting is to search the target document (an ISO up for review) for the key work "Shall". So, if there's more than 10 'shalls' for a 20 page doc., then maybe someone got *excited* and went overboard. In the global environment, you want their support -- not opposition.  I'm around if you need an assist. Stay well!

PR #382, opened 2024-10-10
Application Note consistency

jvdsn:
Right now, this PR does not fully address #381, which is why it is a draft. We need to discuss which "should"s need to be come "must"s


woodbe:
@jvdsn do you expect to do more than this? This looks fine, and anything else we can tag for v3. Just trying to close out what we can before tomorrow.

jvdsn:
@woodbe let's keep this for v3 if that's ok

PR #383, opened 2024-10-10
FMT_MSA.3 SD consistency

jvdsn:
Original comment:

> Application Note 29 states:
> Attributes assigned by the TOE to any parsed SDOs must be described in the Security Target and in operational user guidance.
> However, there is no corresponding guidance EA for FMT_MSA.3 in section 2.4.2.2.
> Either remove the requirement for the operational user guidance from Application Note 29 or include guidance EA.

I added it to the guidance EA, and also made it more explicit in the TSS EA.

### DIFF
--- a/SD_DSC.adoc
+++ b/SD_DSC.adoc
@@ -1383,7 +1383,7 @@ The evaluator shall confirm that the TSS describes the allowed values for each o
 
 ====== AGD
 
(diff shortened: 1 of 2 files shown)


woodbe:
Adjust to note that not all cases may have AGD, note this in the cPP and the SD

slpotte (approved):
The changes look reasonable to me.

Issue #386, opened 2024-10-18
[cPP PUBLIC REVIEW] FCS_STG_EXT.1.1 mismatch


jvdsn:
FCS_STG_EXT.1.1 as defined in the ECD does not match FCS_STG_EXT.1.1 as instantiated in the cPP.

ECD definition:

FCS_STG_EXT.1.1:: The TSF shall provide [assignment: _protection method_] protected storage for asymmetric private keys and [selection: _symmetric keys, persistent secrets, no other keys_].

cPP instantiation:

FCS_STG_EXT.1.1:: The TSF shall provide {empty}[selection: [.underline]#mutable hardware-based, immutable hardware-based, software-based in accordance to FCS_CKM_EXT.3]# protected storage for {empty}[selection: [.underline]#asymmetric private keys, symmetric keys]# and {empty}[selection: [.underline]#persistent secrets, no other keys]#.

(Ignore the formatting for now).

In the ECD, asymmetric private keys is not part of the selection, there is only one selection. In the cPP instantiation there's two selections and asymmetric private keys is part of one of them. I'm inclined to say the cPP instantiation is correct.

jvdsn:
I'm also not sure why asymmetric keys is actually "mandatory". Is the idea that a DSC will always have at least one asymmetric key? I don't think we can change the actual meaning of the SFR now, so the least invasive option is to change the ECD to match the cPP body.

jvdsn:
Maybe it should just be one big selection that includes asymmetric private, symmetric, persistent secrets, and no other keys

woodbe:
The initial discussion was by BSI wanting to be able to do symmetric keys as mandatory, but we ended up not being able to come up with any actual use case where only symmetric keys would be stored, but we moved symmetric to mandatory.

Take a look at the pull request. Basically instead of having the weird first selection I remove it to just make them mandatory and then leave in the one at the end.

Also removes the underlined part because that was an assignment moving to a selection, so it shouldn't be there.

PR #390, opened 2024-10-19
Sfr formatting

### LINKED ISSUE #378: [cPP PUBLIC REVIEW] [atsec11] SFR formatting
> In many SFRs, the key word "selection" and "assignment" should be bolded.

We have discussed this before in the context of the ECD. We didn't extend that effort to the full cPP, is that something we want to target for the 2.0 release?

### LINKED ISSUE #388: [cPP PUBLIC REVIEW] FDP_SDC.2.1 mismatch
FDP_SDC.2.1 as instantiated in the cPP does not match FDP_SDC.2.1 as defined in CC22

CC22 definition:

The TSF shall ensure the confidentiality of the [selection: all user data, the following user
data [assignment: list of user data]] according to [assignment: data characteristics] while
it is stored under the control of the TSF.

cPP instantiation:

FDP_SDC.2.1:: The TSF shall ensure the confidentiality of the [.underline]#[authorization data and the following user data [assignment: _list of internally and externally stored SDEs_]#] according to {empty}[assignment: [.underline]#_SDEs identified in the confidential SDE list attribute of an SDO_#] while it is stored under the control of the TSF.

(Ignore the formatting for now).

"authorization data" is not defined in CC22. I see two options:
- Refinement:
- Define "authorization data and the following user data [...]" to actually be the instantiation of the assignment in the selection in the CC22 definition. Then it would become something like "shall ensure the confidentiality of the [the following user data: authorization data, [assignment: SDEs identified in the confidential SDE list attribute of an SDO]]" essentially replacing the assignment with another assignment.

### LINKED ISSUE #389: [cPP PUBLIC REVIEW] FPT_TST.1.1 mismatch
FPT_TST.1.1 as instantiated in the cPP does not match FPT_TST.1.1 as defined in CC22

FPT_TST.1.1 as defined in CC22 includes ": [assignment: _list of self-tests run by the TSF_]"; I think this is a simple fix to just copy the assignment over, but I'm open to suggestions.

jvdsn:
First, update the SFR formatting conventions section as discussed
during the meeting (the most substantial change being that assignment
-> selection refinements won't be bolded completely).

Then, for each SFR, the formatting is updated to match this convention.

The only exceptions are FCS_STG_EXT.1.1, FDP_SDC.2.1, FPT_TST.1.1, due
to other issues.

Note that this PR includes the commits from #385 for my convenience. It should be merged after #385 is merged.

The remaining SFRs are discussed in #386 , #388 , #389 

Closes #378 
Closes #388 
Closes #389

### DIFF
--- a/cPP_DSC.adoc
+++ b/cPP_DSC.adoc
@@ -1453,7 +1453,7 @@ _Application Note {counter:remark_count}_:: _When an SDE is a key then it is als
 
 FDP_SDC.2 Stored data confidentiality with dedicated method
 
-FDP_SDC.2.1:: The TSF shall ensure the confidentiality of the [.underline]#[authorization data and the following user data [assignment: _list of internally and externally stored SDEs_]#] according to {empty}[assignment: [.underline]#_SDEs identified in the confidential SDE list attribute of an SDO_#] while it is stored under the control of the TSF.
+FDP_SDC.2.1:: The TSF shall ensure the confidentiality of [.line-through]#the# {empty}[[.underline]#the following user data [_authorization data, [assignment: SDEs identified in the confidential SDE list attribute of an SDO]_]#] according to [assignment: *SDEs identified in the confidential SDE list attribute of an SDO*] while it is stored under the control of the TSF.
 
 FDP_SDC.2.2:: The TSF shall ensure the confidentiality of the user data specified in FDP_SDC.2.1 without user intervention.
 
--- a/cPP_DSC.adoc
+++ b/cPP_DSC.adoc
@@ -1855,7 +1855,7 @@ _Application Note {counter:remark_count}_:: _The TSF receives authorization from
 
 FPT_TST.1 TSF Testing
 
-FPT_TST.1.1:: The TSF shall run a suite of self tests *during power-on start-up*, {empty}[selection: [.underline]#periodically during normal operation, at the request of the authorized user, *at no other condition,* at the conditions [_assignment: conditions under which self test should occur_]]# to demonstrate the correct operation of [the TSF].
+FPT_TST.1.1:: The TSF shall run a suite of the following self-tests *during power-on start-up* {empty}[selection: [.underline]#periodically during normal operation, at the request of the authorized user, *at no other condition,* at the conditions [_assignment: conditions under which self-test should occur_]]# to demonstrate the correct operation of {empty}[[.underline]#the TSF#]: [assignment: _list of self-tests run by the TSF_].
 
 FPT_TST.1.2:: The TSF shall provide authorized users with the capability to verify the integrity of {empty}[[.underline]#TSF data#].
 
--- a/cPP_DSC.adoc
+++ b/cPP_DSC.adoc
@@ -1380,7 +1380,7 @@ FDP_ACF.1.2:: The TSF shall enforce the following rules to determine if an opera
 * _[assignment: rules automatically enforced by the TSF that conditionally permit certain subject-object-operation actions based on subject security attributes, object security attributes, or other conditions]_
 * _[selection: [assignment: any configurable rules or parameters that can be modified to affect the behavior of the Access Control SFP], no configurable rules]]_].
 
-FDP_ACF.1.3:: The TSF shall explicitly authorize access of subjects to objects based on the following additional rules: [_assignment: rules, based on security attributes, that explicitly authorize access of subjects to objects_].
+FDP_ACF.1.3:: The TSF shall explicitly authorize access of subjects to objects based on the following additional rules: [assignment: _rules, based on security attributes, that explicitly authorize access of subjects to objects_].
 
 FDP_ACF.1.4:: The TSF shall explicitly deny access of subjects to objects based on the following additional rules: [_client applications can only access their own data, [*assignment*: rules, based on security attributes, that explicitly deny access of subjects to objects]_].
 
--- a/cPP_DSC.adoc
+++ b/cPP_DSC.adoc
@@ -1374,7 +1374,7 @@ FDP_ACF.1.2:: The TSF shall enforce the following rules to determine if an opera
 
 FDP_ACF.1.3:: The TSF shall explicitly authorize access of subjects to objects based on the following additional rules: [assignment: _rules, based on security attributes, that explicitly authorize access of subjects to objects_].
 
-FDP_ACF.1.4:: The TSF shall explicitly deny access of subjects to objects based on the following additional rules: [_client applications can only access their own data, [*assignment*: rules, based on security attributes, that explicitly deny access of subjects to objects]_].
+FDP_ACF.1.4:: The TSF shall explicitly deny access of subjects to objects based on the following additional rules: [_client applications can only access their own data, [assignment: rules, based on security attributes, that explicitly deny access of subjects to objects]_].
 
 _Application Note {counter:remark_count}_:: _The expectation of this SFR is that the reader is given sufficient information to determine, for each object controlled by the TOE, the operations that can be performed on it based on the subject attempting to perform the operation, and whether this is conditional based on attribute values or any other circumstances._
 +
--- a/cPP_DSC.adoc
+++ b/cPP_DSC.adoc
@@ -589,22 +589,18 @@ The conventions used in descriptions of the SFRs are as follows:
 * Unaltered SFRs are stated in the form used in [CC2] or their extended component definition (ECD);
 * Refinement made in the PP: the added/removed text is indicated with *bold text*/[.line-through]#strikethroughs#. When text is substituted (i.e. some text is added in place of some other text, which is then deleted), only the added text is included;
 
-Note that a refinement is also used to indicate cases where the PP replaces an assignment defined for an SFR in [CC2] and replaces it with a selection;
-
-* Selections:
-
-** Wholly or partially completed in the PP: the selection values (i.e. the selection values adopted in the PP or the remaining selection values available for the ST) are indicated with [.underline]#underlined text#;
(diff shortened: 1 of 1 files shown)


cPP_DSC.adoc
```diff
@@ -589,22 +589,18 @@ The conventions used in descriptions of the SFRs are as follows:
 * Unaltered SFRs are stated in the form used in [CC2] or their extended component definition (ECD);
 * Refinement made in the PP: the added/removed text is indicated with *bold text*/[.line-through]#strikethroughs#. When text is substituted (i.e. some text is added in place of some other text, which is then deleted), only the added text is included;
 
-Note that a refinement is also used to indicate cases where the PP replaces an assignment defined for an SFR in [CC2] and replaces it with a selection;
```
jvdsn, line 592: This sentence was removed because it's explained better below ("If the original SFR defined an assignment operation which was completed by the PP author by redefining it as a selection operation ...")

cPP_DSC.adoc
```diff
@@ -1372,23 +1368,21 @@ _The following are the SFP-relevant security attributes that are associated with
...
-* _{empty}[selection: [.underline]#[assignment: any configurable rules or parameters that can be modified to affect the behavior of the Access Control SFP], no configurable rules]#_].
+* _[selection: [assignment: any configurable rules or parameters that can be modified to affect the behavior of the Access Control SFP], no configurable rules]]_].
 
 FDP_ACF.1.3:: The TSF shall explicitly authorize access of subjects to objects based on the following additional rules: [_assignment: rules, based on security attributes, that explicitly authorize access of subjects to objects_].
 
-FDP_ACF.1.4:: The TSF shall explicitly deny access of subjects to objects based on the following additional rules: [_client applications can only access their own data, [assignment: rules, based on security attributes, that explicitly deny access of subjects to objects]_].
+FDP_ACF.1.4:: The TSF shall explicitly deny access of subjects to objects based on the following additional rules: [_client applications can only access their own data, [*assignment*: rules, based on security attributes, that explicitly deny access of subjects to objects]_].
```
woodbe, line 1385: This assignment should not have bolding, this is literally exactly what is in the CC:2022 SFR

cPP_DSC.adoc
```diff
@@ -1372,23 +1368,21 @@ _The following are the SFP-relevant security attributes that are associated with
...
 * _[assignment: rules automatically enforced by the TSF that always prohibit certain subject-object-operation actions]_
 * _[assignment: rules automatically enforced by the TSF that always permit certain subject-object-operation actions]_
 * _[assignment: rules automatically enforced by the TSF that conditionally permit certain subject-object-operation actions based on subject security attributes, object security attributes, or other conditions]_
-* _{empty}[selection: [.underline]#[assignment: any configurable rules or parameters that can be modified to affect the behavior of the Access Control SFP], no configurable rules]#_].
+* _[selection: [assignment: any configurable rules or parameters that can be modified to affect the behavior of the Access Control SFP], no configurable rules]]_].
 
 FDP_ACF.1.3:: The TSF shall explicitly authorize access of subjects to objects based on the following additional rules: [_assignment: rules, based on security attributes, that explicitly authorize access of subjects to objects_].
```
woodbe, line 1383: The assignment here should actually not be italics, the _ should be move to "rules". This looks like an earlier editing error.

woodbe (changes requested):

cPP_DSC.adoc
```diff
@@ -653,12 +649,12 @@ Extended SFRs (i.e. those SFRs that are not defined in [CC2]) are identified by
...
 FCS_CKM.1 Cryptographic Key Generation
 
-FCS_CKM.1.1:: The TSF shall generate cryptographic keys in accordance with a specified cryptographic key generation algorithm *corresponding to [.underline]#[selection:*#
+FCS_CKM.1.1:: The TSF shall generate cryptographic keys in accordance with a specified cryptographic key generation algorithm *corresponding to* [*selection*:
 +
-* [.underline]#*Asymmetric keys generated in accordance with FCS_CKM.1/AKG,*#
-* [.underline]#*Symmetric keys generated in accordance with FCS_CKM.1/SKG,*#
-* [.underline]#*Derived keys generated in accordance with FCS_CKM.5*#
-* [.underline]#*Derived keys generated in accordance with FCS_CKM_EXT.8*#
+* _Asymmetric keys generated in accordance with FCS_CKM.1/AKG,_
```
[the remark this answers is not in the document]
jvdsn: Yes, that is correct. What we are doing here is not completing a selection, it is replacing an assignment with a selection (which kind-of completes the assignment, but also performs a refinement). There are very few places where we actually wholly/partially complete a selection (since that would be a selection defined in CC22/ECD).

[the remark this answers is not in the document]
woodbe: ok ,just wanted to make sure this is how we are going. I don't have an issue with it, but wanted to be certain we agreed.

cPP_DSC.adoc
jvdsn: It's a bit pedantic, but not exactly. What we do here is we partially complete the assignment and replace it with another assignment (which is actually not really defined by our conventions). This is in CC:2022:

[assignment: rules, based on security attributes, that explicitly deny access of subjects to objects]

this is what we do (formatting removed):

[client applications can only access their own data, [assignment: rules, based on security attributes, that explicitly deny access of subjects to objects]].

Maybe we should rewrite this as a refinement?

The TSF shall explicitly deny access of subjects to objects*, such that client applications can only access their own data, and* based on the following additional rules: [assignment: _rules, based on security attributes, that explicitly deny access of subjects to objects_]

cPP_DSC.adoc
jvdsn: Technically this whole thing is a completion of an assignment, with some nested selections and assignments within to make it more complicated for us :). We don't have any explicit rules in that case (only the case where an assignment becomes a selection) so I just used the general rule of italicizing everything.

cPP_DSC.adoc
woodbe: The way I would read it is that we are actually making a partially filled out assignment, so the "client applications can only access their own data" should be underlined, but then leave the rest along as those are just the normal text.

So if you want to say we are replacing the assignment with itself, OK, but then since we are filling out the assignment, then that needs to be underlined according to the convention, since we are forcing something in that assignment.

So if the assignment has to be bold, then the other part needs underlining, and then I think it is fine. This is really one of those cases where we should probably have just written the SFR the way we want it instead of trying to be so generic and then having to further refine our own EXT requirement.

cPP_DSC.adoc
```diff
@@ -1235,7 +1231,7 @@ _In the third selection for FCS_RBG.1.3, "on demand" means, that a TOE presents
 
 FCS_OTV_EXT.1 One-Time Value
 
-FCS_OTV_EXT.1.1:: The TSF shall perform cryptographic one-time value generation for [*selection*: _algorithm or mode_] using the output of a {empty}[*selection*: _random bit generator as defined in FCS_RBG.1, deterministic IV construction, [assignment: OTV construction method]_] and sizes of length that meet the following: [*selection*: _list of standards_].
+FCS_OTV_EXT.1.1:: The TSF shall perform cryptographic one-time value generation for [*selection*: _algorithm or mode_] using the output of a {empty}[selection: _random bit generator as defined in FCS_RBG.1, deterministic IV construction, [assignment: OTV construction method]_] and sizes of length that meet the following: [*selection*: _list of standards_].
```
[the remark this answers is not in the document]
woodbe: ok, that's fine (doesn't show while looking at the read through since it misses the able part).

cPP_DSC.adoc
woodbe: Sorry, this was on FDP_ACF.1.3, not the previous modification. The assignment there should not be in italics, that is the earlier editing error. This requirement is identical to CC:2022.

cPP_DSC.adoc
jvdsn: This is not an EXT requirement, so we didn't write this one. In any case, I'm fine removing the bold from the *assignment* to close this out.

woodbe: The original CC:2022 SFR is:

>The TSF shall explicitly deny access of subjects to objects based on the following additional rules: [assignment: _rules, based on security attributes, that explicitly deny access of subjects to objects_].

Clearly we placed a specific thing you have to do "client applications can only access their own data" into the original assignment, which needs to be underlined, as that would be part of the "partially filled" things. 

The rest though is identical to the CC:2022 text, which to me means just leave it along. If you want to indicate it has been modified because of the partially filled part, that's OK, I can get that, but the "client applications can only access their own data" needs to be underlined since we are requiring that to be part of the assignment. 

I read that as specifying one of the rules that has to be there. You MAY have additional rules, but you have to have that one, and so that is partially filled out.

jvdsn: @woodbe I agree with you, except that partially filled must be italic, not underlined, since we're partially filling an assignment.

woodbe (approved):

jvdsn:
@woodbe do the FDP_SDC.2.1 and FPT_TST.1.1 changes also look good to you? (I probably should've created separate PRs for those, sorry about that).

woodbe:
On looking back at it I think the one thing is that the FDP_ACF.1.4 should still have the underline for "client applications can only access their own data" but otherwise I think this is all good. The FDP_SDC.2 and FPT_TST.1 changes are fine.

jvdsn:
> On looking back at it I think the one thing is that the FDP_ACF.1.4 should still have the underline for "client applications can only access their own data" but otherwise I think this is all good. The FDP_SDC.2 and FPT_TST.1 changes are fine.

@woodbe FDP_ACF.1.4 is an assignment, not a selection in CC22. Per our conventions, "• Assignment wholly or partially completed in the PP: indicated with italicized text;", why would we use underline?

woodbe:
because in my head all the partials were done that way. Go ahead and merge it

PR #396, opened 2025-01-14
Clarify key agreement key derivation algorithms

jvdsn:
NIST prefers the term KDA for the algorithms specified in SP 800-56Cr2, so it doesn't hurt to follow them to avoid confusion. (https://github.com/usnistgov/ACVP-Server/issues/160)

Additionally, add the "-1S" prefix to KDA-HASH to clarify that this is really still a OneStep KDA. Finally, some input information was missing (e.g., output length, fixed information also apply to hash-based OneStep KDA).


jvdsn:
This commit also explicitly specifies the MACs allowed in KDA-MAC-1S, in preparation of a different commit. Per SP 800-56Cr2, only HMAC or KMAC are allowed (CMAC is not allowed). I added HMAC-SHA-1, HMAC-SHA-256, HMAC-SHA-512 which are listed in KDA-KMAC-2S, but as far as I'm concerned any HMAC with a hash function listed in FCS_COP.1/Hash shold be fine. I also added KMAC128 and KMAC256, which I think is quite reasonable.

PR #397, opened 2025-01-14
Remove part of the application note for FCS_CKM.5

jvdsn:
I don't understand why the application not restricts the algorithms used for various purposes. All of these algorithms should be equally suitable in terms of _output_ values. In terms of _input_ values, there's certainly a distinction (e.g., the SP 800-108r1 KDFs are supposed to take a key as input whereas the SP 800-56Cr2 KDAs take a shared secret as input). Maybe that distinction could be clarified in the app notes.

Of course if someone can explain the reasoning behind this app note to me, we can abandon this PR.


jvdsn:
Now in #422.

PR #398, opened 2025-01-14
Update SP 800-108r1 algorithms in FCS_CKM.5

jvdsn:
- The inputs should match SP 800-108r1
- The references seemed to be messed up, I would expect ISO/IEC 11770-6:2016 and SP 800-108r1 to be on the same line since those are equivalent references
- KDF-KMAC should be moved up to be with the other SP 800-108r1 KDFs


jvdsn:
Also, the output key length of 192 was added for KDF-KMAC. Even though the PRF is KMAC128 or KMAC256, it's perfectly fine to use KMAC256 to derive a 192-bit output key since KMAC256 provides 256 bits of security strength.

jvdsn:
Now in #422.

PR #399, opened 2025-01-15
Merge FCS_COP.1/KeyedHash and FCS_COP.1/CMAC

jvdsn:
HMAC and CMAC serve the same security goals, so it would make sense to have only one Security Functional Requirement. This simplifies our cPP and allows more flexibility in the future (e.g. if we wanted to add other MACs such as GMAC).

There are some tricky areas though:
- FCS_CKM_EXT.8 depended on FCS_COP.1/KeyedHash. However since that's our SFR, we can just remove the dependency. This doesn't really change the security properties, since by definition you'll need to implement HMAC in order to have a properly functioning PBKDF2 implementation.
- The KDF-MAC-2S referred to FCS_COP.1/KeyedHash, this has now been changed to an explicit list of allowed MACs.
- In the SD, the number of test groups and test cases was different for HMAC and CMAC (remember this is based on ACVP, which does not consistently provide the same number of test groups and test cases). Right now the new minimum for FCS_COP.1/MAC is the minimum of both SFRs (1 test group of 8 test cases).

Note this includes some commits from #396 so this cannot be merged until that PR is merged.

### DIFF
--- a/SD_DSC.adoc
+++ b/SD_DSC.adoc
@@ -1962,7 +1962,7 @@ If [.underline]#Concatenated key5# is selected for any key derivation function, 
 
 [conditional] If a KDF is used to form a KEK, the evaluator shall ensure that the TSS includes a description of the key derivation function and shall verify the key derivation uses an approved derivation mode and key expansion algorithm according to SP 800-108 Rev. 1 or SP 800-56C Rev. 2.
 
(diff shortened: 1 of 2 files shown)


jvdsn:
Obsoleted by Crypto Catalog updates.

PR #408, opened 2025-03-27
Align FCS_CKM_EXT.7 with CWG

jvdsn:
- ECDH-NIST and ECDH-BPC were merged to ECDH
- ECDH-Ed was rewritten to clarify it involves Montgomery curves, not Edwards curves
- ECIES, KAS-KDF, KAS-KDF-KEK, and KAS-KDF-XOR were removed

I agree with these technical changes made by the CWG, however I believe the removal of ECIES could be contentious.

- In the extended component definition for this SFR, we use "assignment" instead of "selection" since there is nothing to select at that point.
- editorial/formatting

### DIFF
--- a/cPP_DSC.adoc
+++ b/cPP_DSC.adoc
@@ -693,9 +693,11 @@ _Some selections allow assignment of "some value that does not contain any CSP."
 
 FCS_CKM_EXT.7 Cryptographic Key Agreement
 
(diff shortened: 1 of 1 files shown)


slpotte (approved):
EDIES is more of a protocol using multiple cryptographic algorithms rather than a standalone algorithm.  A PP author can always add an SFR for this.

jvdsn:
> EDIES is more of a protocol using multiple cryptographic algorithms rather than a standalone algorithm. A PP author can always add an SFR for this.

@slpotte I agree, I'm not particularly for or against including it, I just thought it was interesting it was removed and wanted to point it out to the wider group.

woodbe (approved):

Issue #377, opened 2024-10-07
Review for "hidden requirements"


woodbe:
On the final review it seemed that there may be some "hidden requirements" buried in the various app notes.

Cases where the note says something like "if they say this, then they need to include some other SFR" but where that SFR isn't mentioned in the requirement itself, nor is it a dependency for that requirement, or linked in any other way. Most of these I think we handled, but on the final review a few showed up when looked at. These were not caught in the v1.0 review or the majority of the v2.0 review, so this is an explicit issue to do such a review for the subsequent version.

jvdsn:
I read through most app notes (though I didn't focus on this issue specifically) and only found this:

FTP_CCMP_EXT.1: `_Inclusion of this SFR requires inclusion of AES-CCM in FCS_COP.1/AEAD._`

FTP_GCMP_EXT.1: `_Inclusion of this SFR requires inclusion of AES-GCM in FCS_COP.1/AEAD._

This could be resolved by rewriting the SFR text of FTP_CCMP_EXT.1 (and FTP_GCMP_EXT.1) as follows:

> The TSF shall implement CCMP using AES in CCM mode, as specified in FCS_COP.1/AEAD, and key size [...]

However that SFR is an extended SFR and the way it is defined right now does not allow such an insertion. Of course we could update the extended component definitions, but probably not for v2.0.

PR #409, opened 2025-03-27
Align FCS_COP.1/Hash with CWG

jvdsn:
The Crypto Working Group did not make any technical changes to this SFR.

- In the application note, "The hash selection should be consistent" was changed to "The hash selection shall be consistent"
- In the application note, "SHA-1 should not be used in applications in which collision resistance is needed" was changed to "SHA-1 shall not be used in applications in which collision resistance is needed"
- editorial/formatting

This is potentially relevant to #377 and #381

### DIFF
--- a/SD_DSC.adoc
+++ b/SD_DSC.adoc
@@ -607,7 +607,7 @@ There are no KMD evaluation activities for this component.
 
 ==== Cryptographic Operation (FCS_COP)
 
-===== FCS_COP.1/Hash Cryptographic Operation (Hashing) 
+===== FCS_COP.1/Hash Cryptographic Operation - Hashing
 
 ====== TSS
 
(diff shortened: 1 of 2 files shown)


slpotte (approved):
These changes look good.

However, I didn't see the changes to the application note.  Would your suggested changes in the application note qualify as "hidden requirements"?

jvdsn:
> However, I didn't see the changes to the application note. Would your suggested changes in the application note qualify as "hidden requirements"?

@slpotte I think it's a fair question. Yes, right now the "shall" statements in the application notes would probably be hidden requirements. I think what @woodbe was pointing out in #377 was particularly the requirements that say "if X is selected then Y must be included". Perhaps these items should be added to the SFR itself as opposed to the application notes.

woodbe (approved):

PR #412, opened 2025-03-27
Align FCS_COP.1/SigVer with CWG

jvdsn:
- Hash functions for KCDSA and EC-KCDSA were updated
- LMS and HSS were separated in two separate rows
- Additional selections were defined for LMS and HSS 
- XMSS and XMSS^MT^ were separated in two separate rows
- Additional selections were defined for XMSS and XMSS^MT^

I don't have any opinion on the KCDSA and EC-KCDSA changes; I agree with the LMS, HSS, XMSS, and XMSS^MT^ changes made by the CWG.

- Consistency with FCS_COP.1/SigGen
- The application note "FIPS 186-5 allows the use of SHAKE128 and SHAKE256. Implementations must use the correct number of bits of output as specified in FIPS 186-5." seems to be a note we added ourselves or was removed from the Crypto Catalog. I believe it is still useful so I kept it but changed "must" to "shall" per #381.
- editorial/formatting

### DIFF
--- a/SD_DSC.adoc
+++ b/SD_DSC.adoc
@@ -722,7 +722,7 @@ The evaluator shall be able to provide a rationale as to the sufficiency of the 
 
 There are no KMD evaluation activities for this component.
 
-===== FCS_COP.1/SigVer Cryptographic Operation (Signature Verification)
+===== FCS_COP.1/SigVer Cryptographic Operation - Signature Verification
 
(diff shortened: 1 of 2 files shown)


slpotte (approved):
Changes look good.

woodbe (approved):

PR #413, opened 2025-03-27
Align FCS_COP.1/SKC with CWG

jvdsn:
- Add CAM-CFB, CAM-OFB, CAM-CTR

I don't have any opinion on these technical changes made by the CWG, though I'm curious why the same additions weren't made for AES.

### DIFF
--- a/SD_DSC.adoc
+++ b/SD_DSC.adoc
@@ -759,7 +759,7 @@ The evaluator shall be able to provide a rationale as to the sufficiency of the 
 
 There are no KMD evaluation activities for this component.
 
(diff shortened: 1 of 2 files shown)


slpotte (approved):
Although I prefer not to see must or shall in the application note as this looks like hidden requirements, the changes look good.

jvdsn:
@slpotte I agree this is a prime candidate for cleanup per #377. I just added a commit which may look a bit silly (in terms of the SFR text) but removes the "hidden requirement" from the app note. Let me know what you think.

woodbe (approved):

PR #415, opened 2025-03-27
Align FCS_OTV_EXT.1 with CWG

jvdsn:
The Crypto Working Group did not make any technical changes to this SFR.

- In the extended component definition for this SFR, we use "assignment" instead of "selection" since there is nothing to select at that point.
- editorial/formatting


slpotte (approved):
Changes look good.

woodbe (approved):

PR #416, opened 2025-03-27
Align FCS_CKM.1/AKG with CWG

jvdsn:
- Added (3072, 256) option for KCDSA
- LMS and HSS were separated in two separate rows
- Additional selections were defined for LMS and HSS 
- XMSS and XMSS^MT^ were separated in two separate rows
- Additional selections were defined for XMSS and XMSS^MT^
- More application notes regarding when functionality is claimed

I don't have any opinion on the KCDSA changes; I agree with the LMS, HSS, XMSS, and XMSS^MT^ changes made by the CWG. The application notes may be contentious according to #377.

### DIFF
--- a/SD_DSC.adoc
+++ b/SD_DSC.adoc
@@ -1891,7 +1891,7 @@ There are no KMD evaluation activities for this component.
 
 ==== Cryptographic Key Generation (FCS_CKM)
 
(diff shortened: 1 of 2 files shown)


slpotte (approved):
Changes look good.

woodbe (approved):

PR #421, opened 2025-03-27
Align FCS_COP.1/KeyWrap with CWG

jvdsn:
- Added CCM
- Added GCM

I agree with the addition of CCM and GCM, it will reduce the number of references in our cPP (previously we referred to both /KeyWrap and /AEAD when keys were encrypted, now the reference to /AEAD can be removed). However, I don't understand the restriction that the deterministic IV construction method must be used. 

- Made the application note consistent with FCS_COP.1/KeyEncap
- editorial/formatting

### DIFF
--- a/SD_DSC.adoc
+++ b/SD_DSC.adoc
@@ -2190,7 +2190,7 @@ The evaluator shall be able to provide a rationale as to the sufficiency of the 
 
 There are no KMD evaluation activities for this component.
 
-===== FCS_COP.1/KeyWrap Cryptographic Operation (Key Wrap)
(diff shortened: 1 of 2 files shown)


slpotte (approved):
The changes look good.

woodbe (approved):

jvdsn:
@slpotte @robertmclemons could you clarify why for GCM there is this restriction: "IV length must be equal to 96 bits; **the deterministic IV construction method [SP800-38D, Section 8.2.1] must be used**; the MAC length t must be one of the values 96, 104, 112, 120, and 128 bits."

PR #423, opened 2025-03-28
Align FCS_CKM_EXT.8 with CWG

jvdsn:
- Add SHA-3 hash functions as the PRF
- Change the salt length selection to an assignment
- (Unsure) add 512-bit output key length

I agree with these technical changes made by the CWG.

- In the extended component definition for this SFR, get rid of the unnecessary brackets surrounding HMAC and the standard.
- Still allow any output length larger than 128 bits as we agreed upon earlier


slpotte (approved):
The changes all look reasonable.

woodbe (approved):

PR #424, opened 2025-03-28
Align FCS_CKM_EXT.3 with CWG

jvdsn:
I don't think the Crypto Working Group made technical changes to this SFR, though the list of operations in FCS_CKM_EXT.3.1 differs from what we had in the DSC cPP. I think it's more likely the DSC iTC did something custom.

- Move the following application notes into the SFR, as we did previously:
> If “key encapsulation” is selected, FCS_COP.1/KeyEncap must be claimed.
> If “key wrapping” is selected, FCS_COP.1/KeyWrap must be claimed.
> If “key encryption” is selected, FCS_COP.1/SKC or FCS_COP.1/AEAD must be claimed.

Note that FCS_COP.1/SKC was added to key encryption, I'm not sure if this was a change by the CWG or if the DSC iTC decided at some point it was not suitable for key encryption. If the latter, we can remove it.
- editorial/formatting

### DIFF
--- a/SD_DSC.adoc
+++ b/SD_DSC.adoc
@@ -2018,7 +2018,7 @@ The evaluator shall examine the KMD to ensure that:
 
 ====== TSS
 
-The evaluator shall verify that for each key selected cryptograhpic method the ST contains the corresponding SFR. The evaluator shall ensure the key access operations which utilize the specified cryptographic methods. Where multiple methods may be selected, the evaluator shall verify that each method's usage is described.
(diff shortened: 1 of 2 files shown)


slpotte (approved):
SKC lacks the authentication/integrity protection component of the other methods.  Perhaps it's here for backwards compatibility with algorithms already deployed?

slpotte (approved):
Changes look reasonable.

jvdsn:
@slpotte I agree that SKC lacks authentication which is why I'm hesitant to allow it for key encryption. Backwards compatibility is a fair point, and considering I don't fully understand the reasons we have it, I'm fine keeping it to avoid breaking things

woodbe (approved):

PR #442, opened 2025-04-29
Add ML-KEM and ML-DSA support (working)

### LINKED ISSUE #441: [cPP PUBLIC REVIEW] NIAP #14

**Comment Type.**
C

ML-KEM and ML-DSA should be included in the list of algorithms for FCS_CKM.1/AKG and FCS_COP.1/SigGen, SigVer

**Describe alternatives you've considered**

**Additional context**

jvdsn:
This is intended to be released in a future version of the cPP.

The Crypto Catalog seems to have two SFRs for KeyEncap and KeyDecap separately. I don't think this is necessary, so for now I kept one SFR.


slpotte (approved):
We will probably only add Encap to catalog as well.  Thanks for feedback.

PR #443, opened 2025-05-02
Remove unnecessary references to AEAD

jvdsn:
With FCS_COP.1/KeyWrap now including GCM and CCM, it is no longer required to specify /AEAD in most instances.

The only open instance is FCS_CKM_EXT.3, where /AEAD is used in conjunction with /SKC.


slpotte (approved):
This SFR in the catalog has new wording which removes the references to the SFRs from the selections.

woodbe (approved):

PR #444, opened 2025-05-05
Merge FPT_MFW_EXT.2 and FPT_MFW_EXT.3

jvdsn:
Both FPT_MFW_EXT.2 and FPT_MFW_EXT.3 are selection-based SFRs that are only included when "mutable" firmware is used. There are no situations in which one is selected without the other. Consequently, it is more appropriate to merge these SFRs into one SFR that covers the requirements for both.

Additionally, FPT_MFW_EXT.2 seems to unnecessarily pull in requirements from the Root of Trust for Measurements. This seems like a scope creep that should be avoided, as some TOEs do not want/need to provide these measurements.

For this reason, the new SFR contains two core requirements: integrity checking on boot (i.e., before firmware execution), and integrity checking of firmware updates. The new SFR also forces the ST author to explicitly select the algorithms used to verify the integrity (and authenticity).

It is expected that the list of allowed cryptographic algorithms here should be reduced. While hash or MAC algorithms are sometimes used to verify the integrity of firmware prior to execution, it is almsot never appropriate to use symmetric (i.e., hash, MAC, AEAD) cryptography to verify firmware updates.

### DIFF
--- a/SD_DSC.adoc
+++ b/SD_DSC.adoc
@@ -2340,33 +2340,11 @@ There are no KMD evaluation activities for this component.
 
 ==== Mutable/Immutable Firmware (Extended - FPT_MFW_EXT)
 
-===== FPT_MFW_EXT.2 Basic Firmware Integrity
+===== FPT_MFW_EXT.2 Firmware Integrity
 
 ====== TSS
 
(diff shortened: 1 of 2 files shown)


woodbe:
This looks good, but we want to review the attestation SFRs and check that these are things that can be reported in that or if other changes should be made to accommodate removal of that from these.

jvdsn:
Moved to #475

PR #445, opened 2025-05-05
Make FCS_CKM_EXT.7 selection-based

jvdsn:
Currently, FCS_CKM_EXT.7 is a mandatory SFR, which unnecessarily
restricts the DSCs that claim compliance to this PP. The SFR can be
made selection-based, since it is included in the selection for
FTP_ITE_EXT.1.


jvdsn:
Note: must be merged after #443

slpotte (approved):
The DSC is free to determine the proper location of these SFRs.

woodbe (approved):

Issue #428, opened 2025-04-29
[cPP PUBLIC REVIEW] NIAP #1


jvdsn:
**What is the location of the document where the comment is made?**
Section 2.2.1, "If the SDOs have been erased (e.g. due to a tamper response), then a factory reset may not be possible."

**Comment Type.**
A

**What is the comment? Please describe.**
Section 2.2.1, 2nd full paragraph: In the 1st sentence, was the erasure of SDOs meant, or SDEs? The context seems to be SDEs.

**Describe the proposed solution/edit**

**Describe alternatives you've considered**

**Additional context**
Clarity

PR #447, opened 2025-06-04
Address NIAP  comment 1

jvdsn:
See #428

I agree SDEs makes more sense from the context, but I don't know why this sentence was written so I may be wrong


woodbe (approved):

woodbe:
In the context of the surrounding text, SDE makes more sense. Overall, probably doesn't matter so much. The problem is that this should probably be talking about them as SDOs  and not SDEs anyway since an SDE could contain only a single SDE.

Simplest change now though is this.

PR #449, opened 2025-06-04
Address NIAP comment 5

### LINKED ISSUE #432: [cPP PUBLIC REVIEW] NIAP #5
Section 2.2.2, Table 2

**Comment Type.**
S

For Parse, mention SDE so that the reader doesn't get to Parse and wonder where the SDEs came from.

**Describe alternatives you've considered**

**Additional context**
Clarity

jvdsn:
See #432

### DIFF
--- a/cPP_DSC.adoc
+++ b/cPP_DSC.adoc
@@ -220,7 +220,7 @@ DSCs provide seven core security services to a platform as illustrated in <<Core
 |Description
 
 |Parse
-|The DSC shall ingest pre-installed keys, credentials, tokens, attributes, etc. from trusted components or services external to its boundary either across a secured channel or in a manner that the objects are protected for use only by the DSC.
+|The DSC shall ingest SDEs (e.g., pre-installed keys, credentials, tokens) and attributes from trusted components or services external to its boundary either across a secured channel or in a manner that the objects are protected for use only by the DSC.
 
 |Provision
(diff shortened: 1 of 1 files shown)


woodbe:
Should this be SDOs and not SDEs because all the other descriptions focus on an SDO ? That would seem to be more consistent.

jvdsn:
According to Figure 4, the SDEs are ingested together with the attributes, and then the Provision service converts them into SDOs. So during parsing it's still just SDEs.

woodbe:
Sort of, most of the time. Sometimes an SDO may be imported in whole, even if that isn't clear in the diagram. 

I think what I would say is more like this:

>The DSC shall ingest pre-installed keys, credentials, tokens as SDEs and where applicable, with attributes to form SDOs, from trusted components or services external to its boundary either across a secured channel or in a manner that the objects are protected for use only by the DSC.

This way it specifies how these external things are used to build the foundation of SDEs and SDOs

jvdsn:
But even if SDOs are imported as a whole, would that not be the Provision service? The table currently says:

> The DSC shall create SDOs from parsed or generated SDEs and attributes using
binding mechanisms to apply integrity protection to the SDEs together with their
attributes.

woodbe:
@woodbe will update the PR with the proposed text and send for review

woodbe (approved):

woodbe:
@jvdsn take a look, just made the edit from the suggested comment.

jvdsn:
Looks good!

Issue #434, opened 2025-04-29
[cPP PUBLIC REVIEW] NIAP #7


jvdsn:
**What is the location of the document where the comment is made?**

**Comment Type.**
S

**What is the comment? Please describe.**
The PP needs to be reviewed to identify and remove requirements embedded anywhere other than in SFR or SAR text. In particular, requirements must not be embedded in narrative text (e.g., page 15, “the ST author is expected to…”) or within application notes for SFRs.

**Describe the proposed solution/edit**

**Describe alternatives you've considered**

**Additional context**

jvdsn:
Seems identical to #377

PR #452, opened 2025-06-04
Address NIAP comment 7

jvdsn:
See #434


woodbe (approved):

woodbe:
I'm not sure that the specific issue is really a concern given this isn't even in the requirements, but I'm fine with it.

jvdsn:
Yeah I agree, this seems very minor.

Issue #436, opened 2025-04-29
[cPP PUBLIC REVIEW] NIAP #9


jvdsn:
**What is the location of the document where the comment is made?**
Application Note 3.

**Comment Type.**
S

**What is the comment? Please describe.**
This is an example of a “shall” in an application note. Do not use “shall” in application notes. Application notes must be solely informative. In this case, the rewording is easy: “The ST author completes the first assignment to list all..”

**Describe the proposed solution/edit**

**Describe alternatives you've considered**

**Additional context**

jvdsn:
Seems related to #381

woodbe:
Reviewed and completed in several pull requests

PR #453, opened 2025-06-04
Remove shall from some Application Notes

jvdsn:
This is to address #381 and #436, but does not resolve all instances.


woodbe:
For FCS_CKM_EXT.8. couldn't we change the assignment for the salt to say:

[**assignment**: _equal to or greater than 128 as provided by FCS_OTV_EXT.1_]

Or something similar of there is a better way to write it? Then the statement about dependencies in the app note doesn't apply as the linkage is explicit in the SFR. I know that isn't written into the catalog, but I think it should be.

slpotte:
You could.  However, there's a possibility that FCS_OTV_EXT.1 will go away.

woodbe (approved):

PR #458, opened 2025-06-13
Appnote7

woodbe:
app note 7 only

### DIFF
--- a/cPP_DSC.adoc
+++ b/cPP_DSC.adoc
@@ -842,9 +842,9 @@ NIST SP 800-208 [parameters]
 
 |===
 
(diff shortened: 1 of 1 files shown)


woodbe (commented):
@jvdsn The last two points of the app note seem like nuances of the specific implementation, but I don't know. 

The last one actually seems to literally just be a statement that is taken from the spec, I don't think that needs to be here.

The second to last though seems like there could be options, like different hashed COULD be used, but here are expected to be the same. I have no clue about the algorithms themselves, but is this how they are supposed to work, or is this a limiting option required here by the WG?

I would think to get rid of them, leaving only the first three as notes since I think the last few are actually unnecessary or are hidden requirements that need to be specified in the table.

jvdsn:
@woodbe this seems to be covered by #453 already.

PR #456, opened 2025-06-13
Appnoteupdate

woodbe:
This is a review of the first for App notes related to #435 and #436. A small heading has been added to the Cryptographic section heading noting about ensuring commensurate protection of keys in the DSC. This will be taken out of the remaining app notes.

This covers the first 5 app notes.

PR #459, opened 2025-06-13
Appnote8

### LINKED ISSUE #457: FCS_COP.1/Hash App note #5
So the App note here really is a minefield for the "special" words. Interestingly this wasn't called out by NIAP, but:

> SHA-1 **may** be used as a general hash function and for the following applications: generating and verifying hash-based message authentication codes (HMACs), key derivation functions (KDFs), and random bit/number generation. SHA-1 **may** also be used for verifying old digital signatures and time stamps, if this is explicitly allowed by the application domain. SHA-1 **shall not** be used in applications in which collision resistance is needed.

These are clearly "hidden" requirements in the app note detailing where and how the hash functions can be used.

The real issue here is that SHA1 can't be used where collision resistance is needed and may be used where it is reading old data. This could be done in the SD certainly, ensuring that it isn't used in inappropriate places, but what this almost screams to me is actually adding a table for this but adding something about acceptable uses, which would seem to require an extended SFR. That or iterate it and modify the one for SHA1.

woodbe:
This is to fix App note 8. This is a lot like app note 7 #458 in terms of what is there.

The main issue is DSA which is like #457 and the SHA1. With limited usage, this seems to be more than an app note

### DIFF
--- a/cPP_DSC.adoc
+++ b/cPP_DSC.adoc
@@ -956,9 +956,7 @@ NIST SP 800-208 [parameters]
 
 |===
 
-_Application Note {counter:remark_count}_:: _FIPS 186-5 allows the use of SHAKE128 and SHAKE256._
-+
-_DSA may only be used to verify signatures generated prior to the implementation date of FIPS 186-5._
+_Application Note {counter:remark_count}_::_DSA may only be used for verifying old digital signatures and time stamps, if this is explicitly allowed by the application domain._
 +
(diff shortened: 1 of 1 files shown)


jvdsn:
#453

woodbe:
@jvdsn I think we should still delete the other two lines based on the changes I have here, so if you integrate those into #453 then I think we are good.

slpotte (approved):
Looks good.

woodbe:
Look at SHA1 change in #456 and see about adjusting DSA line similarly.

woodbe:
@jvdsn can you check the update here to make sure I got it right? I think this is it, and I removed the SHAKE line while I was at it.

jvdsn:
Either way is fine by me

jvdsn (approved):

PR #460, opened 2025-06-13
Appnote8 & 10

woodbe:
For SigVer we have the same issues as in app note 7 with the last three lines looking like they already exist in the specs and don't need to be here.

The DSA though has the same issue as SHA1 in that it is only allowed for certain uses. This needs to be in the SFR somehow and not part of the app note. It would seem to be very hidden if stated in the SD.

For FCS_OTV_EXT.1 I know that it is possible this SFR may just disappear, but we have it now, so we have to deal with it. Most of it though is a description of what salts, nonces and IVs are. I don't think we need that in the app note. So I am pulling everything except the last line, I think that is actually best.

### DIFF
--- a/cPP_DSC.adoc
+++ b/cPP_DSC.adoc
@@ -956,15 +956,7 @@ NIST SP 800-208 [parameters]
 
 |===
 
-_Application Note {counter:remark_count}_:: _FIPS 186-5 allows the use of SHAKE128 and SHAKE256._
-+
-_DSA may only be used to verify signatures generated prior to the implementation date of FIPS 186-5._
-+
-_For LMS, HSS, XMSS, and XMSS^MT^, the key sizes do not represent the expected security strength. All key sizes given here correspond to an expected security strength of 128 bits, per NIST SP 800-208._
-+
-_For HSS and XMSS^MT^ the same hash or XOF function shall be used at each level. Within each level, the same Winternitz parameter shall be used but can be different for each level. For HSS, within each level, the same tree height shall be used but can be different for each level._
-+
-_Per Section 5.4 of NIST FIPS 204, the collision strength of the hash function selected for the pre-hash of HashML-DSA must be at least the strength of the signature algorithm._
+_Application Note {counter:remark_count}_:: _DSA may only be used for verifying old digital signatures and time stamps, if this is explicitly allowed by the application domain._
 
 ==== FCS_COP.1/SKC Cryptographic Operation - Symmetric-Key Cryptography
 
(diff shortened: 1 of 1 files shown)


cPP_DSC.adoc
```diff
@@ -956,14 +956,10 @@ NIST SP 800-208 [parameters]
 
 |===
 
-_Application Note {counter:remark_count}_:: _FIPS 186-5 allows the use of SHAKE128 and SHAKE256. Implementations shall use the correct number of bits of output as specified in FIPS 186-5._
+_Application Note {counter:remark_count}_:: _FIPS 186-5 allows the use of SHAKE128 and SHAKE256._
```
jvdsn: This one is already covered by #453

cPP_DSC.adoc
```diff
@@ -956,14 +956,10 @@ NIST SP 800-208 [parameters]
...
 +
 _DSA may only be used to verify signatures generated prior to the implementation date of FIPS 186-5._
 +
-_The TOE may contain a public key which is integrity protected (e.g., in hardware), in which case the FDP_ITC.1 and FDP_ITC.2 dependencies do not apply. In this case, no dependencies may be chosen._
-+
```
jvdsn: Agree with removing this.

cPP_DSC.adoc
```diff
@@ -956,14 +956,10 @@ NIST SP 800-208 [parameters]
...
-_For signature verifications, private keys are not necessary, so there are no dependencies required for generating or destroying cryptographic keys._
```
jvdsn: Is this removal warranted because the app note is intended for PP authors not ST authors?

cPP_DSC.adoc
```diff
@@ -1258,15 +1254,7 @@ The following table provides the allowed choices for completion of the selection
...
-+
-_Initialization Vectors (IVs) help protect against attacks which depend on the reuse of static keys. Certain encryption modes often require IVs. They should be randomly generated in a nonpredictable way, cannot be sequential, and cannot repeat._
-+
-_Each algorithm and mode have varying guidance on the lengths of the salts, nonces, and initialization vectors used therein. Please consult the referenced standards documents for the appropriate guidance for each._
+_Application Note {counter:remark_count}_:: _Each algorithm and mode have varying guidance on the lengths of the salts, nonces, and initialization vectors used therein. Please consult the referenced standards documents for the appropriate guidance for each._
```
jvdsn, line 1249: I liked the background explanation in the app note but you are right it's probably better to just refer to the standards.

jvdsn (changes requested):

slpotte (approved):

jvdsn (approved):

PR #461, opened 2025-06-13
Appnote11-15

woodbe:
This only changes note 13 to remove "expected" language

PR #462, opened 2025-06-13
Appnote16-20

woodbe:
App note 18 seems like a hidden dependency in that it is stating that specific SDEs are also subject to FCS_CKM.6. I actually don't think that is necessary, since the two requirements are not mutually exclusive. Maybe this should just say something about FCS_CKM.6 is an acceptable method for destruction when appropriate.

App note 19 similarly seems to be quite prescriptive in the types of required data. These could be added as selections or even suggested assignments in the SFR itself. I would probably recommend selections, not the assignment.

### DIFF
--- a/cPP_DSC.adoc
+++ b/cPP_DSC.adoc
@@ -1385,7 +1385,7 @@ FDP_ITC_EXT.1.4:: The TSF shall bind SDEs to security attributes using [assignme
 
 _Application Note {counter:remark_count}_:: _The way the TSF checks the integrity of the SDE depends on the method of importation. For example, the encrypted data channel may provide data integrity as part of its service._
 +
-_When a TSF parses an SDE, it should generate security attributes and create an SDO by binding the security attributes to the SDE._
+_When a TSF parses an SDE, an SDO will be created by binding it to generated security attributes._
 
 ==== FDP_ITC_EXT.2 Parsing of SDOs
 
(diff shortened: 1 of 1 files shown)


cPP_DSC.adoc
```diff
@@ -1345,11 +1345,11 @@ FDP_ACF.1.4:: The TSF shall explicitly deny access of subjects to objects based
...
+_Subject-object-operation combinations that are always permitted and those that are conditionally permitted, combined with rules about the remaining combinations such as blacket prohibition of access is sufficient._
 +
-_FDP_ACF.1.3 and FDP_ACF.1.4 allow the ST author to optionally specify override conditions to resolve otherwise contradictory Access Control SFP rules. For example, the rule "S.Admin may always modify the SDO.Conf attribute of any OB.P_SDO or OB.T_SDO object" may be overridden by a statement in FDP_ACF.1.4 that identifies any particular SDO objects as non-modifiable regardless of subject authorizations._
+_FDP_ACF.1.3 and FDP_ACF.1.4 allow for optional override conditions to resolve otherwise contradictory Access Control SFP rules. For example, the rule "S.Admin may always modify the SDO.Conf attribute of any OB.P_SDO or OB.T_SDO object" may be overridden by a statement in FDP_ACF.1.4 that identifies any particular SDO objects as non-modifiable regardless of subject authorizations._
 +
```
jvdsn, line 1353: Everything above this line is covered by #461

jvdsn (changes requested):

woodbe:
Make sure this is done AFTER the pull request before it and then MAKE SURE it doesn't re-add those changes that were fixed in the earlier update.

jvdsn (approved):

PR #463, opened 2025-06-16
App notes 21-25

woodbe:
This is to update the app notes 21-25. App note 25 has a potential issue that I have removed, but don't know where it shoud be re-added, or if it even can.

>The ST should also indicate the location of those policies, and how the TOE protects the integrity of those policies

I don't know that this is something that we can have here. This requirement is beyong FIA_UAU.6 and specifies information about location and integrity, which seem to be FDP or FTP something or other.

### DIFF
--- a/SD_DSC.adoc
+++ b/SD_DSC.adoc
@@ -1254,7 +1254,7 @@ There are no KMD evaluation activities for this component.
 
 ====== TSS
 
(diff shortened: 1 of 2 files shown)


slpotte (approved):
Changes look good.  The bit about the policy may go in the SD?

PR #465, opened 2025-06-16
Appnote41-45

woodbe:
In FCS_RBG.2 I have removed this. 

> The TSF protects the integrity and confidentiality of the entropy it receives from the external entropy source

This is not part of this requirement, there isn't anything about protecting the entropy that is received, so I don't see how to put it in here as it is a hidden requirement.


slpotte (approved):

woodbe:
Look at adding this as sensitive data, under an FDP or FTP or some definition as to what constitutes sensitive data.

Probably don't need to be specific about whether the source is internal or external, any entropy data should be considered sensitive and be treated as secure

Issue #55, opened 2022-12-20
CCDB GE/013 - Section 2.1.3.3.3


woodbe:
Para 85

### Comment:
Some words are written with capital letters randomly.

### Priority: 2

woodbe:
will fix in v1.1

PR #468, opened 2025-07-15
AppNotes55-59

woodbe:
For #55:

The last sentence is a little confusing, especially the last part with the FDP_FRS_EXT.1.1 because that is an open assignment, so there technically isn't an option for "no actions or conditions". Obviously this is something someone can fill out, but it's quite strange to say that this is required if someone doesn't provide information in an assignment as opposed to a selection.

For 58 and 59, the last paragraph in each talk specifically about cryypto requirements or almost implementation details for each of the methods for doing this, showing crypto requirements. While I wouldn't dispute that should be how to do implement these, given that it isn't specified in the SFR itself, there seems to be a hidden requirement about how to answer that SFR, which doesn't have any selections/assignments. It would seem that maybe the SFR should have this information specified for the crypto implementation if they are being specified so clearly here.

### DIFF
--- a/cPP_DSC.adoc
+++ b/cPP_DSC.adoc
@@ -2935,9 +2935,9 @@ FPT_FLS.1/FW:: Failure with Preservation of Secure State (Firmware)
 
 FPT_FLS.1.1/FW:: The TSF shall preserve a secure state when the following types of *firmware* failures occur: [_authenticity violation, integrity violation, rollback violation_].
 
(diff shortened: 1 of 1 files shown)


jvdsn:
For 55: according to our ECD, FDP_FRS_EXT.1 actually depends on FDP_FRS_EXT.2, so would it even be valid to select FDP_FRS_EXT.1 without FDP_FRS_EXT.2...?

For 58 and 59, it can be pulled into the SFRs, like I did in #444, but it's a pretty major change, since you'd have to add a selection. Probably not suitable for a 2.1.

jvdsn (approved):
The actual changes in this PR are good.

Issue #474, opened 2025-09-25
FDP_FRS_EXT.1 and FDP_FRS_EXT.2


jvdsn:
FDP_FRS_EXT.1 is currently a mandatory SFR but has the option "no actions or conditions" to allow the DSC to **not** support a factory reset. This does not make sense.

FMT_SMF.1 currently includes "Reset TOE to factory state for FDP_FRS_EXT.1" as a mandatory management function, but evidently the DSC is allowed to **not** support a factory reset. Again, this does not make sense.

The solution seems to be as follows:
- Move "Reset TOE to factory state for FDP_FRS_EXT.1" to the selection of management functions.
- Make FDP_FRS_EXT.1 selection-based and remove the "no actions or conditions option".

Now, it becomes straightforward to merge FDP_FRS_EXT.1 and FDP_FRS_EXT.2 into one SFR.

PR #477, opened 2025-10-13
Reduce number of mandatory SFRs

jvdsn:
This PR contains 5 commits with the main goal to reduce the number of mandatory SFRs (just the number of SFRs in general) to allow more flexibility with the cPP.

#### Merging FPT_MFW_EXT.2 and FPT_MFW_EXT.3 is done as in #444, so that PR will be closed

Both FPT_MFW_EXT.2 and FPT_MFW_EXT.3 are selection-based SFRs that are only included when "mutable" firmware is used. There are no situations in which one is selected without the other. Consequently, it is more appropriate to merge these SFRs into one SFR that covers the requirements for both.

Additionally, FPT_MFW_EXT.2 seems to unnecessarily pull in requirements from the Root of Trust for Measurements. This seems like a scope creep that should be avoided, as some TOEs do not want/need to provide these measurements.

For this reason, the new SFR contains two core requirements: integrity checking on boot (i.e., before firmware execution), and integrity checking of firmware updates. The new SFR also forces the ST author to explicitly select the algorithms used to verify the integrity (and authenticity).

It is expected that the list of allowed cryptographic algorithms here should be reduced. While hash or MAC algorithms are sometimes used to verify the integrity of firmware prior to execution, it is almsot never appropriate to use symmetric (i.e., hash, MAC, AEAD) cryptography to verify firmware updates.

#### FPT_RPL.1/Rollback is renamed to FPT_RPL.1/FW

This is mainly for consistency with FPT_FLS.1/FW. However, I also want to point out that FPT_FLS.1/FW has significant overlap with FPT_RPL.1/Rollback and FPT_MFW_EXT.2. I prefer to avoid overlap.

#### FPT_TST.1 test are made more generic

Currently this SFR focuses heavily on TSF / TSF data integrity, but that's already covered by a lot of SFRs elsewhere (e.g., FPT_MFW_EXT.2, FPT_FLS.1/FW, some of the RoT SFRs). It makes more sense to focus on the cryptographic algorithm self-tests.

#### Root of Trust becomes a core SFR

The FPT_PRO_EXT.1, FPT_ROT_EXT.1, and FPT_ROT_EXT.2 SFRs were merged into one mandatory SFR called FPT_ROT_EXT.1. This SFR provides a few selections for the confidentiality and integrity of data. The idea is that a lot of the crypto SFRs can become selection-based based on these selections (not implemented yet). That would allow more flexibility for TOEs that don't support certain cryptographic algorithms. This SFR also specifies 3 RoT services that could be implemented (selection-based): Measurement, Reporting, and Verification.

Measurement and Reporting are simply FPT_PRO_EXT.2 and FPT_ROT_EXT.3 but renamed (and slightly edited). Verification is new, as it seems to be a useful service that can make use of the signature algorithms.

Finally, one thing to discuss here is whether or not FCS_STG_EXT.1 serves any purpose. It was already covered by the old Root of Trust for Storage, and there's many more SFRs that cover SDO protection, import, and export. Maybe it should be merged?

#### FDP_FRS_EXT.1 and FDP_FRS_EXT.2 are merged and selection-based

As explained in #474:

FDP_FRS_EXT.1 is currently a mandatory SFR but has the option "no actions or conditions" to allow the DSC **to** not support a factory reset. This does not make sense.

FMT_SMF.1 currently includes "Reset TOE to factory state for FDP_FRS_EXT.1" as a mandatory management function, but evidently the DSC is allowed to **not** support a factory reset. Again, this does not make sense.

The solution seems to be as follows:
- Move "Reset TOE to factory state for FDP_FRS_EXT.1" to the selection of management functions.
- Make FDP_FRS_EXT.1 selection-based and remove the "no actions or conditions option".

Now, it becomes straightforward to merge FDP_FRS_EXT.1 and FDP_FRS_EXT.2 into one SFR.

### DIFF
--- a/SD_DSC.adoc
+++ b/SD_DSC.adoc
@@ -1598,7 +1598,7 @@ There are no KMD evaluation activities for this component.
 
 ====== TSS
 
-The evaluator shall examine the TSS and other vendor documentation and ensure they describe the methods used to verify integrity of the TSF and TSF data, both automatic checks and if supported, checks requested by the user. If more than the power-on check is selected, then each condition for the self test shall be examined.
+The evaluator shall examine the TSS and other vendor documentation and ensure they describe the methods used to demonstrate the correct operation of the TSF, verify the integrity of the TSF and its data, both automatic checks and if supported, checks requested by the user. The TSS shall also describe how the TSF responds to self-test failures. Each condition for the self-tests shall be described and examined.
 
 ====== AGD
 
@@ -1606,11 +1606,9 @@ The evaluator shall examine the operational guidance to ensure it provides autho
 
 ====== Test
 
-Test 1: The evaluator shall verify that the DSC performs an integrity check of all TSF, including data, as well as performing KATs for those functions. The evaluator shall verify failures using malformed known answer test data (for example, unexpected input or output values).
+Test 1: For each implemented self-test, the evaluator shall cause the self-test to fail and verify the TSF responds as described in the TSS.
 
-Test 2: The evaluator shall ensure that when an integrity check failure occurs specific to failing KATs and failure to verify the integrity of the TSF, the TOE will prevent any further processing of the current TSF and user data. 
-
-Test 3 [conditional]: For each condition where a self test may be initiated, the evaluator shall verify that the self tests are run as specified.
+Test 2 [conditional]: For each condition where a self-test may be initiated, the evaluator shall verify that the self-tests are performed as specified.
 
 ====== KMD
 
--- a/SD_DSC.adoc
+++ b/SD_DSC.adoc
@@ -2437,7 +2437,7 @@ There are no KMD evaluation activities for this component.
 
 ==== Replay Detection (FPT_RPL)
 
-===== FPT_RPL.1/Rollback Replay Detection (Rollback)
+===== FPT_RPL.1/FW Replay Detection (Firmware)
 
 ====== TSS
 
@@ -2451,7 +2451,7 @@ There are no AGD evaluation activities for this component.
 
 The evaluator shall repeat the following tests to cover all allowed firmware verification mechanisms as described in the TSS. For example, if the firmware verification mechanism replaces an entire partition or subset of the DSC scope containing many separate code files, the evaluator does not need to repeat the test for each individual file.
 
-Test 1: The evaluator shall attempt to execute an earlier instance or build of the software (as determined by the vendor). The evaluator shall verify that this attempt fails by checking the version identifiers or cryptographic hashes of the firmware against those previously recorded and checking that the values do not correspond to an unauthorized build.
+Test 1: The evaluator shall attempt to execute an earlier instance or build of the firmware (as determined by the vendor). The evaluator shall verify that this attempt fails by checking the version identifiers or cryptographic hashes of the firmware against those previously recorded and checking that the values do not correspond to an unauthorized build.
 
 Test 2: The evaluator shall attempt to execute a current or later instance and shall verify that the firmware execution succeeds.
 
--- a/SD_DSC.adoc
+++ b/SD_DSC.adoc
@@ -985,38 +985,6 @@ Testing for this SFR is performed as part of FCS_CKM.2.
 
 There are no KMD evaluation activities for this component.
 
-==== Factory Reset (Extended - FDP_FRS_EXT)
-
-===== FDP_FRS_EXT.1 Factory Reset
-
-====== TSS
-
-The evaluator shall examine the TSS to determine that it describes each of the conditions which may lead to a factory reset.
-
-====== AGD
-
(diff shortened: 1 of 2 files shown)


SD_DSC.adoc
```diff
@@ -2397,59 +2342,47 @@ There are no KMD evaluation activities for this component.
...
+The TOE guarantees the authenticity of the firmware using the identity of the Guarantor. This prevents impersonating a Guarantor when sending firmware to a device or modifying the firmware in transit.
 
-The evaluator shall examine the TSS to ensure it describes the methods and identities used to verify integrity and authenticity of the firmware. The TSS shall identify the Guarantor and how to verify its identity.
+*Test 1: Verify Authentic Firmware*
 
-====== AGD
+The evaluator shall trigger the TOE to load and evaluate the integrity and authenticity of authentic firmware before execution, according the methods described in the TSS. The evaluator shall ensure that the TOE provides a clear indication of the success of the evaluation to consider the test a 'Pass', otherwise, the test is a 'Fail'.
 
-There are no guidance activities for this component.
+*Test 2: Verify Unauthentic Firmware*
```
woodbe, line 2363: Inauthentic? Technically either is fine, but "in" is more common for this than "un"

SD_DSC.adoc
```diff
@@ -2397,59 +2342,47 @@ There are no KMD evaluation activities for this component.
...
+The evaluator shall trigger the TOE to load and evaluate the integrity and authenticity of the deliberately modified firmware before execution, according the methods described in the TSS. The evaluator shall ensure that the TOE provides a clear indication of the failure of the evaluation to consider the test a 'Pass', otherwise, the test is a 'Fail'.
 
-*Test 1: Verify Authentic Firmware*
+*Test 3: Verify Authentic Firmware Updates*
 
-The evaluator shall trigger the TOE to load and evaluate the authenticity of authentic firmware according the methods described in the TSS. The evaluator shall ensure that the TOE provides a clear indication of the success of the evaluation to consider the test a 'Pass', otherwise, the test is a 'Fail'.
+The evaluator shall trigger the TOE to load and evaluate the integrity and authenticity of an authentic firmware update, according the methods described in the TSS. The evaluator shall ensure that the TOE provides a clear indication of the success of the evaluation to consider the test a 'Pass', otherwise, the test is a 'Fail'.
 
-*Test 2: Verify Unauthentic Firmware*
+*Test 4: Verify Unauthentic Firmware Updates*
```
woodbe, line 2373: Same as above. Not major either way, just a comment.

SD_DSC.adoc
```diff
@@ -2397,59 +2342,47 @@ There are no KMD evaluation activities for this component.
...
+The evaluator shall trigger the TOE to load and evaluate the authenticity of the deliberately modified firmware update, according the methods described in the TSS. The evaluator shall ensure that the TOE provides a clear indication of the failure of the evaluation to consider the test a 'Pass', otherwise, the test is a 'Fail'.
 
 ====== KMD
 
 There are no KMD evaluation activities for this component.
 
 ==== Replay Detection (FPT_RPL)
 
-===== FPT_RPL.1/Rollback Replay Detection (Rollback)
+===== FPT_RPL.1/FW Replay Detection (Firmware)
```
woodbe, line 2385: Should we write out "Firmware" instead of "FW"? Not a major issue either way, but give many are written out, maybe it should be?

cPP_DSC.adoc
```diff
@@ -2899,21 +2860,26 @@ FDP_DAU.1.1/Prove:: The TSF shall provide a capability to generate evidence that
...
 
-_Application Note {counter:remark_count}_:: _Not all DSCs permit factory reset functionality. Those that do are expected to perform a factory reset in a manner that prevents any inadvertent disclosure of security-relevant data that was present on the DSC prior to the factory reset. For DSCs that permit factory reset functionality (as indicated by selection of [.underline]#factory reset the TOE wiping out all non-permanent SDEs and SDOs, as described by FDP_FRS_EXT.2# in FIA_AFL_EXT.1.3, or by [.underline]#no actions or conditions# NOT being selected in FDP_FRS_EXT.1.1), this SFR must be claimed._
+_Application Note {counter:remark_count}_::
+_Not all DSCs permit factory reset functionality. Those that do are expected to perform a factory reset in a manner that prevents any inadvertent disclosure of security-relevant data that was present on the DSC prior to the factory reset. For DSCs that permit factory reset functionality (as indicated by selection of [.underline]#factory reset the TOE wiping out all non-permanent SDEs and SDOs, as described by FDP_FRS_EXT.1# in FIA_AFL_EXT.1.3 or [.underline]#Reset TOE to factory state for FDP_FRS_EXT.1# in FMT_SMF.1), this SFR is claimed._
```
woodbe, line 2880: Is this app note laying a requirement on the user that should be expressed within the SFR itself? I think so, and maybe it requires adjusting how the selections are offered.

woodbe (approved):

PR #479, opened 2025-10-13
Make all FCS_COP SFRs selection-based

jvdsn:
Now that the FDP_* SFRs include selections for the various cryptographic algorithms, move them to the selection-based section. This makes the cPP more flexible in allowing DSCs that don't implement any of the previously mandatory algorithms, ...


woodbe (approved):
This is a large change, but makes sense. I'm OK with moving ahead with it as a starting point for v3 changes and then working on additional updates from there.

PR #483, opened 2025-11-17
v2.1 publication

woodbe:
Final publication of v2.1


jvdsn (approved):
Updated the publication dates to November 17, 2025.

